PR #13, opened 2016-02-02
Cutting material based on feedback

gvwilson:
Cutting:

* lists
* defensive programming
* arrays

Reducing:

* error messages
* reading tabular data
* plotting
* loops
* conditionals
* writing functions
* programming style
* wrap-up

### DIFF
--- a/index.md
+++ b/index.md
@@ -127,8 +127,9 @@ FIXME:
 
 ### Built-in Functions (and Methods) and Help
 
-*   Teaching: 5 min
-*   Exercises: 5 min
+*   Teaching: 10 min
+    *   Include re-running cells and re-running all
+*   Exercises: 10 min
     *   Chain calculations with max and min
     *   Find a useful method using help(str)
     *   Parsons Problem to achieve specific results with string methods
@@ -148,13 +149,6 @@ FIXME:
 
 ### *Coffee: 15 min*
 
-### Pandas Data Frames
-
-*   Teaching: 10 min
-*   Exercises: 10 min
-    *   Import Pandas
-    *   Create and display a data frame
-
 ### Reading Tabular Data
 
 *   Teaching: 5 min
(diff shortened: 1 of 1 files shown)


lexnederbragt:
I hate to see lists go, and running python scripts outside of the notebook, but I see no other way (this is difficult planning!). BTW lists is still mentioned in the 'Learners Will Be Able To...' part.

gvwilson:
Thanks - I've removed mention of lists.

ttimbers:
Some suggestions:

* I would add stressing the importance of run all, and running cells in order under the Running and Quitting Interactively section. I've observed many students who create disorganized code when they learn to code in the notebook (they somehow can remember in their heads which cells to run in which order, but another person would never be able to run their code).

* I am not sure about the value of showing how to create a data frame before importing csv as a data frame. I think most people will do the latter in their workflow and rarely create a data frame from scratch.

* Since it is relatively trivial to export a csv from excel, I think we could skip How tor read data from Excel spreadsheets directly (needs another library). Only showing how to import .csv will hopefully encourage folks to save their own data as a .csv as opposed to .xls.

ttimbers:
Also, can we please do plotting via Pandas, it is much simpler for simple cases.

gvwilson:
Addressed @ttimbers comments in 21acba6203.

PBarmby:
This version looks pretty good; I don't have anything to add. I am really curious to know how this would go with real learners.

gvwilson:
Me too :-)  Shall we try it out some time?

tobyhodges:
I agree - the only way to know what else needs tweaking in the material and timings is by testing it on live subjects. I'm aiming to organise a workshop later this year and could use that as an opportunity to try the lesson. Of course, you'll probably want someone with more SWC experience to do this, and sooner too!

PR #34, opened 2016-03-17
max was being called on a float (non-iterable)

cdw:
Challenge would raise an exception when called. Just adding another value to prevent such.


gvwilson:
Nice catch, and nice exercise - thanks.

PR #41, opened 2016-04-27
Update 15-conditionals.md

rrlove:
Added a processing step to the "processing small files" example. Printing the filename and number of records as the processing step allows learners to immediately see if they got it right; knowledge of len() is already assumed in lesson 4.

### DIFF
--- a/_episodes/15-conditionals.md
+++ b/_episodes/15-conditionals.md
@@ -49,7 +49,7 @@ FIXME: lesson content.
 > for filename in glob.glob('data/*.csv'):
 >     contents = pd.read_csv(filename)
 >     ____:
->         FIXME: what to include as processing step?
+>         print filename,len(contents)
 > ~~~
 > {: .source}
 {: .challenge}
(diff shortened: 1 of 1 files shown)


_episodes/15-conditionals.md
```diff
@@ -49,7 +49,7 @@ FIXME: lesson content.
 > for filename in glob.glob('data/*.csv'):
 >     contents = pd.read_csv(filename)
 >     ____:
->         FIXME: what to include as processing step?
+>         print filename,len(contents)
```
wking: +1 on the change, but we're using [Python][1] [3][2], so `print(filename, len(contents))`.

[1]: https://github.com/swcarpentry/python-novice-gapminder/blob/ca400f197039a50ff280b92bdfd68ba4ad90a47a/_extras/design.md#lists
[2]: https://github.com/swcarpentry/workshop-template/blob/e39365cb9a9bf8186435f92810cc254a2cc914c7/index.html#L620

PR #58, opened 2016-06-15
Instructor Training: Added a function exercise

sshende:
This is an Instructor Training assignment PR. We guide learners
through a multipart, faded example exercise to design a function
that computes - for a given decade - the average GDP for a
country in its specified continent.


gvwilson:
Thanks!

PR #61, opened 2016-06-17
A short instruction on how to load Gapminder data using Pandas library, to the "index.md" file

aliojjati:


gvwilson:
Thanks for the PR - it satisfies the requirements for instructor training - but we can't ask people to do this in `index.md`, since they aren't shown how to start Python until the first episode, or told what `import` means until episode 6.

PR #63, opened 2016-06-18
Added a challenge question in 01-run-quit.md

SeanRGB:
Created a challenge question in 01-run-quit.md asking the user to write some code in a code cell and then try converting the entire cell to markdown. For me this was something that I did when I first tried Jupyter to get a feel for switching between the cell types and as an easy method of commenting out experimental code. 

This is for instructor training checkout. @gvwilson


gvwilson:
Nice!

PR #66, opened 2016-07-01
Inst training

russodanielp:
This is an Instructor Training assignment PR.  Added the beginning of a lesson introducing dictionaries because in the for loop lesson they are mentioned but not explained.


gvwilson:
Thanks for the PR - we've removed the mention of dictionaries from the `for` loop episode.  It's unlikely we're going to be able to include material on dictionaries in this lesson (despite them being my favorite data structure): it's already over-full in most people's opinions, and we'd have to cut even more to make room for dicts.

PR #68, opened 2016-07-22
Fix Python 2 style print statement inside the range for loop

leouieda:
There was a Python 2 print statement inside the loop explaining the range iterator. Since all other prints are functions, I thought this one should be too.


gvwilson:
Thanks!

leouieda:
I'm glad to help!

PR #76, opened 2016-07-23
Reorganization of setup file

biologyguy:
- Restructured in a more natural progressive pattern, from installation of Python, to getting the data, to launching with Jupyter.
- Broken link to gapminder.org fixed.
- A few changes to the actual sentence structure.

### DIFF
--- a/setup.md
+++ b/setup.md
@@ -4,84 +4,52 @@ title: "Setup"
 permalink: /setup/
 ---
 
-## Getting the Data
-
-The data we will be using is taken from the [gapminder](gapminder.org) dataset.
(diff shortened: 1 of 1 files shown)


gvwilson:
Thanks Steve.

biologyguy:
Hey Greg,
Should a link to the setup document be in the 'prerequisites' section of the main lesson page? Students may not dig into the lesson breakdown (which is where the link exists) until the day of.
Just a thought.
-Steve

gvwilson:
Good thought - there should certainly be something on the home page 
pointing people at the setup instructions.  I don't know if 
"prerequisites" is the right place for it - that's intended to be "what 
you need to know before you start" - but we could add it there, or make 
it part of the template.

PR #91, opened 2016-08-09
Pycon as resource

gallingerj:
I added Pycon in next-steps because it is a "community site for
software, workshops, and help."


gvwilson:
Thanks - merging now.

PR #122, opened 2016-10-31
provide solutions to data types and conversion

### LINKED ISSUE #121: working on chapter 03 solutions
Hi there,

Just creating a marker issue to let others know I'm working on solutions for episode 03, "Data Types and Type Conversion". I'll have a pull request out soon.

alee:
- first cut at complete solutions for episode with some minor suggested edits to text
- we could also replace `print(type(foo))` with just `type(foo)` since python will print out the string representation of the returned result if you are following along in the REPL. 
- closes #121

### DIFF
--- a/_episodes/03-types-conversion.md
+++ b/_episodes/03-types-conversion.md
@@ -81,7 +81,7 @@ TypeError: unsupported operand type(s) for -: 'str' and 'str'
 ~~~
 {: .error}
 
-## You can use the "+" and "\*" operators on strings.
+## You can use the "+" and "*" operators on strings.
 
 *   "Adding" character strings concatenates them.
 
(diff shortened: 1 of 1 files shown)


_episodes/03-types-conversion.md
```diff
@@ -81,7 +81,7 @@ TypeError: unsupported operand type(s) for -: 'str' and 'str'
 ~~~
 {: .error}
 
-## Strings can be added and multiplied.
+## You can use the "+" and "\*" operators on strings.
```
ntmoore: Is it a good idea to include the backslash?  I think this might be confusing to a novice.

_episodes/03-types-conversion.md
```diff
@@ -260,21 +260,33 @@ first is 2 and second is 5
 > 4. A lab specimen's age.
 > 5. Current population of a city.
 > 6. Average population of a city over time.
+>
+> > ## Solution
+> >
+> > 1.  _integer_. Number of days since the start of the year.
```
ntmoore: I can think of cases in which these answers are too proscriptive, eg, what plotting, if data is taken at: 
8am on the 3rd, 
noon on the 4th, 
and 6pm on the 4th, 
I can imagine using x-values of 3.333, 4.5, 4.75  

Similarly, a city's population of 141,283 is approximate because of births, dealths, and migration.  Saying the city has a population of 1.41E5 people (using float) doesn't seem wrong to me.

In general, do we need answers for every exercise?  I'm inclined to say no, given te variety of answers some of these questions have.

_episodes/03-types-conversion.md
alee: The backslash was there to escape the `*` so Markdown won't interpret it as beginning an italic section. It won't be rendered in the output. But it doesn't seem to be necessary, so I'm all for its removal :+1:

PR #128, opened 2016-11-21
typo corrections and clarifying phrases

bricegnichols:
minor corrections to describe edit and control modes more clearly


danmichaelo:
The audience for these lessons are probably familiar with keyboard combos and should not be confused by `Pressing the "shift" and the "enter" key together`, but I generally find it slightly clearer to write `Pressing the "enter" key while holding the "shift" key down` (probably because I had to explain this to my mother, who had problems with keyboard combos not always working because she tried to press the keys down at exactly the same moment)

PR #138, opened 2017-01-06
Add fig to `01-run-quit` to show example ipython notebook

katabat:
Added an example figure (see below) of an Jupyter ipython notebook, with the intention of fixing the 

> FIXME:diagram

line in the Running and Quitting lesson

jupyter-example-notebook


katabat:
Hmm, just noticed #124 which seeks to do the same thing... 

#124 gives a better idea of what can be included in notebook, but might this be an better alternative as it includes code that students could immediately try out themselves (ie. retype the contents of the first cell) to replicate the result?

PR #142, opened 2017-01-28
Instructor training. See description for details.

zsmahmood89:
Added very brief mention of '.extend' after the '.append' example in episode 11 on LISTS. Repetition of concept combined with a very intuitive extension of append command might help students internalize the idea of modifying lists.


ntmoore:
I think this is a useful addition.  I think it would be informative with demonstration code with output.  Would you mind proposing typical jupyter notebook instructions and output?

PR #144, opened 2017-02-28
Added solution to problem 1 and 2 in lesson 13 (Looping Over Data Sets) 

cmortiz:
I am submitting this pull request as part of my instructor training checkout.


ntmoore:
Thanks for adding this in.

PR #155, opened 2017-03-29
Say a little more about substrings, slicing, and so on

valtandor:
Also change variable name "element" to "atom_name", as the word "element" is used nearby in another context. See also PR #139 - we may wish to merge in #139 before this one. I'm happy to resolve any conflicts.


valtandor:
I meant to put an "instructor-training" tag on this.

ntmoore:
I don't think this pull was fully accepted into the gapminder lesson.  If you woldn't mind, would you please check to see if the specific additions made it in?

valtandor:
@ntmoore It looks good to me: https://github.com/swcarpentry/python-novice-gapminder/blob/gh-pages/_episodes/02-variables.md

Issue #120, opened 2016-10-29
docker jekyll server


alee:
I set up docker-compose and a simple Dockerfile to make a docker container that can run Jekyll and serve the site. This helps me test my changes to any content before submitting a pull request without having to install jekyll on my local machine. 

If this seems beneficial, I'd be happy to make a pull request with instructions on the README and the associated docker files. This does require that you install [docker engine](https://docs.docker.com/engine/installation/) and [docker compose](https://docs.docker.com/compose/install/).

For a preview of what Docker integration would look like, https://github.com/alee/python-novice-gapminder/commit/7dff37345ae5db5b36767162866a505afc2abe88

ntmoore:
This sounds like it would be useful.  Do you have a running example?

alee:
Hey Nathan,

I just submit a running example that you can test by checking out the referenced PR (or my branch at https://github.com/alee/python-novice-gapminder/tree/docker).

alee:
Closing this for now, something like this should be raised with the styles template group since it applies to all lessons

PR #157, opened 2017-03-29
add docker + docker-compose Jekyll setup

alee:
Using [jekyll/jekyll:pages](https://hub.docker.com/r/jekyll/jekyll/) as
a base image for "as close to Github pages as possible" compliance. See
https://github.com/jekyll/docker/wiki for more details.

This was built with docker-compose version 1.11.2 and
docker version 17.03.1-ce

resolves issue #120

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,6 +8,17 @@ for a rendered version of this material,
 for instructions on formatting, building, and submitting material,
 or run `make` in this directory for a list of helpful commands.
 
+## Run Jekyll in a Docker container
+
+If you download and install the latest versions of [Docker compose](https://docs.docker.com/compose/install/) and
+[Docker engine](https://docs.docker.com/engine/installation/) you can build a Jekyll container to serve this site with
+hot-reloading and incremental rebuilds by running `docker-compose up` and visiting `localhost:4000` in a browser. 
+
+The docker-compose file mounts this root directory into the Jekyll container so you can edit the lesson's Markdown files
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -8,6 +8,17 @@ for a rendered version of this material,
 for instructions on formatting, building, and submitting material,
 or run `make` in this directory for a list of helpful commands.
 
+## Run Jekyll in a Docker container
+
+If you download and install the latest versions of [Docker compose](https://docs.docker.com/compose/install/) and
+[Docker engine](https://docs.docker.com/engine/installation/) you can build a Jekyll container to serve this site with
+hot-reloading and incremental rebuilds by running `docker-compose up` and visiting `localhost:4000` in a browser.
```
ntmoore: I appreciate this addition.  Where specifically should a user run "docker-compose up".  Is thisa Jupyter command, shell command, flie a user should click, etc?

alee: Ah, good point. This is a shell command and also needs some testing on Windows and Mac, since I've only tested it on Linux.. I'll update the readme with more detailed instructions and update this PR.

Ultimately it'd be nice to have a script / tool that runs a Docker "sidecar" Jekyll container. This sidecar could then be attached to any of the software carpentry Jekyll lessons to build & serve the content without having to modify your base system (beyond installing Docker).

alee:
closing, this should be applied to styles instead, see https://github.com/swcarpentry/styles/issues/199

PR #163, opened 2017-04-22
Update 14-writing-functions.md

lizcw:
Replaced last challenge (Encapsulating Data Analysis) with one which builds on previous lessons and provides a demonstrably useful method.
(NB, this is for completion of instructor training)

### DIFF
--- a/_episodes/14-writing-functions.md
+++ b/_episodes/14-writing-functions.md
@@ -77,17 +77,14 @@ print_date(1871, 3, 19)
 ~~~
 {: .output}
 
-*   Via [Twitter](https://twitter.com/minisciencegirl/status/693486088963272705):
-    `()` contains the ingredients for the function
-    while the body contains the recipe.
-
 ## Functions may return a result to their caller using `return`.
 
-*   Use `return ...` to give a value back to the caller.
+*   Use a `return ...` to give result back to the caller.
 *   May occur anywhere in the function.
-*   But functions are easier to understand if `return` occurs:
+*   But functions are easier to understand and debug if `return` occurs:
     *   At the start to handle special cases.
     *   At the very end, with a final result.
+*   Using a variable to capture the return value helps detect errors
 
 ~~~
 def average(values):
@@ -108,7 +105,8 @@ print('average of actual values:', a)
 {: .output}
 
 ~~~
-print('average of empty list:', average([]))
+a = average([])
+print('average of empty list:', a)
 ~~~
 {: .python}
 ~~~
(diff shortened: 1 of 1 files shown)


_episodes/14-writing-functions.md
```diff
@@ -77,17 +77,14 @@ print_date(1871, 3, 19)
 ~~~
 {: .output}
 
-*   Via [Twitter](https://twitter.com/minisciencegirl/status/693486088963272705):
```
ntmoore: I like this analogy and would rather is not be removed from the lesson.  Is there a specific reason you think it should be cut?

lizcw: It doesn't really make sense to me - I actually thought it had been left in by error as the twitter discussion doesn't really reflect the statement it's linked to and is hard to follow.  I don't think calling arguments "ingredients" helps the understanding, in fact, ingredients imply specificity whereas arguments allow for flexibility.   There are references in an earlier lesson to the correct nomenclature (02-variables) so a deviation at this point could create confusion.

ntmoore: If I am making a quiche, the ingredients for the filling are certainly eggs and milk, but also some sort of green vegetable, some sort of cheese, and perhaps some sort of meat.  While the eggs and milk are strictly specified, there is flexibility in other ingredients.  While the analogy of recipes to function definitions is imperfect, I think many people learn by extening known patterns to new concepts, (eg, "Oh, this [new idea] is like [another old idea] that I already understand.  Shayer and Adey call this "bridging" or "transfer" - [see this pdf](http://modeling.asu.edu/modeling/AdeyP-CognitiveAccel2011.pdf) )

Again, I would prefer to keep this as an anecdote at least.  I've noticed the writing in this section is quite terse.  Perhaps if it was presented explicity as an anecdote with explaination...

_episodes/14-writing-functions.md
```diff
@@ -264,82 +262,77 @@ result of call is: None
...
-> 3. How would you generalize this function
->    if you did not know beforehand which specific years occurred as columns in the data?
->    For instance, what if we also had data from years ending in 1 and 9 for each decade?
->    (Hint: use the columns to filter out the ones that correspond to the decade,
->    instead of enumerating them in the code.) 
->
 > > ## Solution
 > >
-> > 1. 
+> > 1. This requires knowledge of the column headings so is of limited re-use
```
ntmoore: I think there are a few problems that are scattered through the lessons that use column headings.  I thinnk we should retain this variation. (Not all of the exercises are required of course.)

k8hertweck:
Thanks so much to for the PR, @lizcw! 

This is a tricky PR to discuss because it includes quite a few changes throughout the length of the module (with the main change described as replacing an exercise). How do you feel about the rest of the changes, @ntmoore? If the rest of it looks OK, here's what I propose as a course of action: Modify the PR to keep the analogy and merge. Open a separate PR specifically about the analogy and solicit some other opinions (perhaps from maintainers of other Python lessons?), so you can decide whether to keep it or develop the language a bit more.

Let me know if there's anything I can do to help!

ntmoore:
I havn't worked through the exercise, but was open to accepting the pr aside from  the deletion of the cooking example.  Final exams right now.  If you want to modify the pr to be two separate pieces, one including the exercise and another starting a discussion of the cooking example, that's fine with me (I'll merge the coding example...)  

I do appreciate your contributions @lizcw!  Community contributions are what make these materials.

lizcw:
As advised, reverting changes. Will open a new branch for replacement challenge example.

PR #167, opened 2017-05-05
ix considered harmful. Use iloc and loc instead

djinnome:
This completely tripped me up in class today when I did not use `index_col='country'` and I called `data.ix[0:10]`  
Because the row indexes were already numbers, `ix` was inclusive at both ends instead of the expected behavior of not including the final index.  This caused extreme confusion and because it was the last episode of the day, my students left confused.
Instead, we should use `iloc` everywhere we want to treat indices like array slices, and `loc` everywhere we want to use index names so that the indexing behavior is isolated to the function not the data type of the index.


_episodes/08-data-frames.md
```diff
@@ -97,11 +108,10 @@ Poland          5338.752143     6557.152776     8006.506993
 ~~~
 {: .output}
 
-In the above code, we discover that **slicing using indexes is inclusive at both
-ends**, which differs from typical Python behavior, where slicing indicates
-everything up to but not including the final index.  However, if we use integers
-when our DataFrame is indexed by something else, slicing follows typical
-pythonic behavior.
+In the above code, we discover that **slicing using loc is inclusive at both
```
rgaiacs: Replace

~~~
In the above code, we discover that **slicing using loc is inclusive at both
~~~

with

~~~
In the above code, we discover that **slicing using `loc` is inclusive at both
~~~

_episodes/08-data-frames.md
```diff
@@ -97,11 +108,10 @@ Poland          5338.752143     6557.152776     8006.506993
...
+ends**, which differs from **slicing using iloc**, where slicing indicates
```
rgaiacs: Replace

~~~
ends**, which differs from **slicing using iloc**, where slicing indicates
~~~

with

~~~
ends**, which differs from **slicing using `iloc`**, where slicing indicates
~~~

djinnome:
I added @rgaiacs suggestion and removed remaining references to `.ix[..., ...]`

PR #173, opened 2017-05-24
Updated links in the document to replace FIXME's with correct links

woodhead99:
My email address for instructor training is greg.woodward@kc.frbres.org


ntmoore:
Thanks for making this correction.

PR #174, opened 2017-05-27
exercise walking learners through process of simulating a dynamical system

davidrpugh:
This is my first contribution.  Feedback would be most appreciated!

### DIFF
--- a/_episodes/14-writing-functions.md
+++ b/_episodes/14-writing-functions.md
@@ -259,14 +259,14 @@ result of call is: None
 >
 > 1. Create a function definition for `print_egg_label()` that will work with the revised program above.  Note, the function's return value will be signifigant. Sample output might be `71.23 large`.
 > 2.  A dirty egg might have a mass of more than 90 grams, and a spoiled or broken egg will probably have a mass that's less than 50 grams.  Modify your `print_egg_label()` function to account for these error conditions. Sample output could be `25 too light, probably spoiled`.
-> 
+>
 {: .challenge}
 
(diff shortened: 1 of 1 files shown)


ntmoore:
I think this is a nice exercise - thanks for contributing!  Part of the fun of this system is the bifurcation plot it produces. (Do I remember the visual is something like this: https://en.wikipedia.org/wiki/Bifurcation_theory#/media/File:Chaosorderchaos.png)  

I think the example would be better if you asked the learner to make a plot and included a few lines in the solution to plot the results.

davidrpugh:
@ntmoore Thanks for the feedback.  I think including a plot as an additional component would be a good idea.  The bifurcation plot is cool but it a bit involved. Should we be worried about cognitive load?

An alternative might be to have learners visually explore sensitive dependence to initial conditions. Basically for a given value of `r` simulate the trajectories for two initial conditions that are `epsilon=1e-6` apart and the measure the absolute deviation between the two trajectories and plot it.  For non-chaotic values of `r`, the two trajectories will converge and the deviation with decay to zero at some rate `lambda<0`; for chaotic values of `r` the two deviation will explode over time at some rate `lambda>0`.  Estimating lambda is akin to estimated the [Lyapunov exponent](https://en.wikipedia.org/wiki/Lyapunov_exponent).  

In the real world demonstrating chaotic behavior for a dynamical system typically boils down to finding parameter values for which at least one dimension of the system has a positive Lyapunov exponent.

Thoughts?

PR #214, opened 2017-07-08
Update 15-scope.md

kcratie:
The questions aren’t appropriate for topic as variable scope is introduced and discussed. The existing questions are on how function calls work and determining errors. Both these questions are better suited to separate sections, e.g., question 1 appears to be better suited in the section writing functions. A section covering introductory debugging techniques would be invaluable to this lesson as it is an inherent part of developing software, and the second question and third objective would go there.


vahtras:

Issue #218, opened 2017-07-16
13-looping-data-sets.md missing from syllabus in rendered version of code


pfujita:
---
Hello!

Looking at the rendered version of this set of lessons, the 13th episode (13-looping-data-sets.md) is not displaying in the schedule/syllabus here: 

https://swcarpentry.github.io/python-novice-gapminder/

I looked at the code, but am not sure why this is happening, so I'm submitting a bug report rather than a fix (sorry!).

Hope this helps!

Pauline
---

delocalizer:
should be fixed by #222

ntmoore:
This seems to be resolved. Thanks for pointing it out and correcting it!

PR #219, opened 2017-07-20
removed byte order mark (BOM) from start of file

delocalizer:
fixes https://github.com/swcarpentry/python-novice-gapminder/issues/218

Looks like the file was created using Notepad or similar editor that saves a BOM at start of the file... which presumably messes with the jekyll parser so that it doesn't see the first three characters as being '---' as required for front matter


delocalizer:
subsumed in #222

PR #220, opened 2017-07-20
fixed formatting

delocalizer:
need space between markdown header tag (##) and the header text so that challenge solutions render properly


delocalizer:
subsumed by #222

PR #238, opened 2017-09-24
Add an introductory note about selecting values

chbrandt:


ntmoore:
I think the changes you propose are fine, but I don't understand why the system says there is a conflict with your revisions.  Would you please check the PR and address the source of conflict?

souravsingh:
Once the conflict is fixed, it should be good to merge.

chbrandt:
@ntmoore @souravsingh 
Looks ok now

Issue #210, opened 2017-06-16
Missing Splitting-Applying-Combining episode


rgaiacs:
The lesson is missing a episode about splitting-applying-combining, see http://pandas.pydata.org/pandas-docs/stable/groupby.html.

ntmoore:
I think this would be a good addition.  It could certainly be a sub section in lesson #8, https://swcarpentry.github.io/python-novice-gapminder/08-data-frames/

alee:
added in 52c2d6d629155489b959a978162e216242fb7e12 

feedback / followup in #276

PR #239, opened 2017-09-24
Add a select-apply-combine section

chbrandt:
An example of grouping and aggregating on dataframe.
This should touch issue #210


ntmoore:
Thanks for adding this vectorizing example.

PR #263, opened 2018-02-01
harmonize example code to reflect the error shown

CanWood:
Modified code to reflect error in "Cannot add numbers and strings"


alee:
Thanks for catching that!

PR #264, opened 2018-02-06
Update 09-plotting.md

narayananr:
The plotting lesson uses "for loops" to get the "year" values from the column names. However, for loop is  introduced later. Instead of for loop, we can use pandas' function to get the year from column names.
`years = data.columns.str.strip('gdpPercap_')
`
instead of the for loop code chunk
`years=[]
for col in data.columns:
     year = col[-4:]
     years.append(year)
`

### DIFF
--- a/_episodes/09-plotting.md
+++ b/_episodes/09-plotting.md
@@ -102,6 +102,9 @@ years = []
 for col in data.columns:
     year = col[-4:]
     years.append(year)
+# For loops is not introduced yet
+# Instead of for loop we can use pandas dataframe functions to get the year.
+years = data.columns.str.strip('gdpPercap_')
 
 # Select two countries' worth of data.
 gdp_australia = data.loc['Australia']


souravsingh:
@narayananr Is there going to be an effect in the learning for newcomers with this change?

ErinBecker:
@narayananr - Thank you for noticing that the for loops are introduced in the lesson after the first time they're used. It's always tricky to keep things in order! 

@souravsingh - I'm not sure whether the best move here is 1) adding the pandas' function here, 2) simplifying the example so that a for loop isn't needed, or 3) moving the introduction to for loops to earlier in the lesson. Maybe your co-maintainers have some thoughts?

narayananr:
thanks for getting back. I found this while teaching the lesson and suggested pandas option as the previous lesson and this is using Pandas. Moving for loops ahead is definitely an option. I think either way is fine.

ntmoore:
I think that we should not move the for loop earlier.  For loops are two lessons later (lesson 12, after lists).  Instead, simplifying this example (to remove for loops and replace with pandas function) seems like a reasonable approach.  pandas and for-loops are both significant cognitive loads for learners - I think its important to keep them separate.

souravsingh:
Sounds good to me.

PR #275, opened 2018-03-05
13, looping data-sets: Terminology & Bugfix

stevenkoenig:
Since regular expressions are not necessarily introduced and globs are not real regular expressions as far as I understand it, I stick to the term used throughout the rest.

Also, I fixed a bug in a challenge (b4419cc) that made the solution always return 0.

### DIFF
--- a/_episodes/13-looping-data-sets.md
+++ b/_episodes/13-looping-data-sets.md
@@ -128,12 +128,12 @@ data/gapminder_gdp_oceania.csv 10039.59564
 > the file that has the fewest records.
 >
 > ~~~
-> import pandas
-> fewest = ____
+> import glob, pandas
+> records = ____
 > for filename in glob.glob('data/*.csv'):
 >     dataframe = pandas.____(filename)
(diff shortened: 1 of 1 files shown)


stevenkoenig:
Updated the PR.

I think that both solutions, yours and mine, are not really clean. Ideally, one would initialize `fewest` to a value indicating that it is unset. Trying it with `None`, I found out that Python's built-in `min` and `max` only work with numbers. I found a way for this to work with `numpy` using 'NaN', but that would need yet another import and `numpy` is not even mentioned before in the material ... :(

So, my [final solution](https://stackoverflow.com/a/19374296) is to initialize `fewest` to `float('Inf')`. IMHO that is the cleanest solution to this problem. It is up to the instructor to point out that one should definitely check for the case that `inf` is the answer and adapt the output accordingly.

alee:
Thanks! Agreed, the most correct way would be yours at the expense of time/space efficiency or explicitly testing and special casing for a sentinel non numeric value.

PR #279, opened 2018-03-07
fix error in solution in episode 6

### LINKED ISSUE #278: error in answers to exercise "Locating the Right Module"
in episode 6, exercise "Locating the Right Module", two of the answers are
```
from random import randrange
random_index = random.randrange(len(bases))
print(bases[random_index])
```
and 
```
from random import randrange
print(bases[random.randrange(len(bases))])
```
in both cases, it should say `randrange` instead of `random.randrange` inside the print function

wikfeldt:
there was an error in the solution to "Locating the right module". fixes #278

### DIFF
--- a/_episodes/06-libraries.md
+++ b/_episodes/06-libraries.md
@@ -172,7 +172,7 @@ cos(pi) is -1.0
 > > ~~~
 > > from random import randrange
 > >
-> > random_index = random.randrange(len(bases))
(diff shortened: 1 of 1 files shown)


souravsingh:
Looks good to me. @alee Are you okay with this?

alee:
Looks good, thanks!

PR #280, opened 2018-03-09
Solved exercise 3 from episode 13: to write a program that plots the …

pakitochus:
Solved exercise 3 from episode 13: to write a program that plots the evolution of the average GDP per capita for each region (also, Instructor-training assignment).

### DIFF
--- a/_episodes/13-looping-data-sets.md
+++ b/_episodes/13-looping-data-sets.md
@@ -157,4 +157,18 @@ data/gapminder_gdp_oceania.csv 10039.59564
 > Write a program that reads in the regional data sets
 > and plots the average GDP per capita for each region over time
 > in a single chart.
+> > ## Solution
+> > ~~~
+> > import pandas 
+> > import matplotlib.pyplot as plt
+> > fig, ax = plt.subplots(1,1)
+> > for filename in glob.glob('data/gapminder_gdp*.csv'):
+> >     dataframe = pandas.read_csv(filename)
+> >     region = filename.split('/')[-1][:-4].split('_')[-1]
+> >     dataframe.mean().plot(ax=ax, label=region)
+> > ax.legend()
+> > plt.show()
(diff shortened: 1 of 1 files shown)


alee:
Overall looks good to me, thanks for the PR! I agree it would be an improvement to unpack Line 167 or at least describe what it's doing (retrieving the region from the `gapminder_gdp_<region>.csv` path). 

You could also simplify it via [rpartition](https://docs.python.org/3/library/stdtypes.html#str.rpartition) e.g.,

```
# extract region from the filename, expected to be in the format 'data/gapminder_gdp_<region>.csv'
region = filename.rpartition('_')[2][:-4] 
```

As a minor note we should  `import glob` at the top.

pakitochus:
I was going to ask about the legibility of the region retrieval line. It's not trivial for a beginner, and to do it in one line might be misleading. I added the suggestion of @alee using `rpartition`, and the readability improves. 

`ax.legend()` was raising a warning in Matplotlib 2.1, so I changed it to `fig.legend()` and the warning stopped. In this case it's not a problem. 

Let me know your thoughts.

alee:
Thanks! I'll verify this solution using pandas + matplotlib later this evening unless someone else gets to it first. `fig.legend()` reported errors in a Jupyter notebook using the [docker data science notebook](https://github.com/jupyter/docker-stacks/tree/master/datascience-notebook) but that may be a version mismatch issue - I'll report back after I investigate further.

alee:
Just tested it on a standalone script and it works fine, so it must be a Jupyter notebook issue. I get a better looking legend using `plt.legend()` instead of `fig.legend()` though..

`fig.legend()`
using fig.legend

`plt.legend()`
using plt.legend

pakitochus:
Thanks for the feedback! I changed the code to `plt.legend()` instead of `fig.legend()` so that the results are more visual.

PR #283, opened 2018-03-11
add solutions to exercises in episode 14

wikfeldt:
this adds new solutions to all exercises in episode 14 except one, and corrects the only pre-existing solution to the last exercise since it had typos, incorrect name of imported pandas library, and used the deprecated `.ix[]` method


souravsingh:
Looks good. Thanks @wikfeldt

PR #289, opened 2018-03-23
make glob more specific

### LINKED ISSUE #131: Addition of file `asia_gdp_per_capita.csv` breaks example
The addition of the file `asia_gdp_per_capita.csv` to `data/` causes the following example to fail because `asia_gdp_per_capita.csv` does not include the column`gdpPercap_1952`.
```
for filename in glob.glob('data/*.csv'):
    data = pandas.read_csv(filename)
    print(filename, data['gdpPercap_1952'].min())
```

Link: [Use glob and for to process batches of files](https://github.com/swcarpentry/python-novice-gapminder/blob/gh-pages/_episodes/13-looping-data-sets.md#use-glob-and-for-to-process-batches-of-files)

alee:
- limit glob pattern to gapminder CSVs with a gdpPercap_1952 field
- fixes #131

### DIFF
--- a/_episodes/13-looping-data-sets.md
+++ b/_episodes/13-looping-data-sets.md
@@ -87,7 +87,7 @@ all PDB files: []
     so that simple patterns will find the right data.
 
 ~~~
-for filename in glob.glob('data/*.csv'):
+for filename in glob.glob('data/gapminder_*.csv'):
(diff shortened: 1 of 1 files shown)


vphill:
This is a pretty straightforward change that fixes #131.  👍

Issue #162, opened 2017-04-20
Plotting introduces loops before loops lesson


amandamiotto:
In https://github.com/swcarpentry/python-novice-gapminder/blob/gh-pages/_episodes/09-plotting.md

You have a loop before you talk about what a loop is.
Also works with lists before explaining lists.

Maybe move this lesson to just after the loops lesson? Not a huge thing but worth mentioning.

vahtras:
I agree completely, my first impression of this lesson is that there is an overall problem with ordering of the sections, like having the most basic parts (lists,  for-loops and if-statements) towards the end of the lesson.

timothylwarren:
I posted issue #296 earlier today before seeing this issue. I suggest an alternative to fix the problem. Short version of my suggestion is that there is no need to use for loops or lists if columns in original csv data file are years not 'gdp_xxx' strings.

eldobbins:
This issue seems to have been resolved.  At least, I cannot find any loops in the current version of Lesson 9.

timothylwarren:
I agree that the issue seems to be resolved in current lesson, which has changed significantly throughout.

alee:
thanks for double checking, @timothylwarren @eldobbins !

PR #294, opened 2018-04-05
Reorder episodes

vahtras:
The only changes in this commit is renaming the files in _episodes
so that the learning curve becomes more logical, with the basic
concepts in the beginning (for, if) rather than the end.

see also issue #162 on

https://github.com/swcarpentry/python-novice-gapminder/issues/162

Please delete the text below before submitting your contribution. 

---

Thanks for contributing! If this contribution is for instructor training, please send an email to checkout@carpentries.org with a link to this contribution so we can record your progress. You’ve completed your contribution step for instructor checkout just by submitting this contribution.  

Please keep in mind that lesson maintainers are volunteers and it may be some time before they can respond to your contribution. Although not all contributions can be incorporated into the lesson materials, we appreciate your time and effort to improve the curriculum.  If you have any questions about the lesson maintenance process or would like to volunteer your time as a contribution reviewer, please contact Kate Hertweck (k8hertweck@gmail.com).  

---


gvwilson:
Thanks for the PR @vahtras and for the ping @ntmoore - yes, the current order seems strange from a "fundamentals first" point of view, but as Guzdial and others have shown with their media-first computing research, introducing applications first is more engaging and produces higher retention. Many people who teach R (both for the Carpentries and in regular classrooms) use a similar approach: they start by loading data into data frames, doing straightforward statistical calculations, and plotting, and only then move on to things like `if` and `for`, and it seems to be more of a ramp and less of a cliff for many learners.

```
@inproceedings{Guzd2013,
  doi = {10.1145/2493394.2493397},
  year = {2013},
  publisher = {Association for Computing Machinery ({ACM})},
  author = {Mark Guzdial},
  title = {Exploring Hypotheses about Media Computation},
  booktitle = {Proc.\ 2013 International Computing Education Research Conference ({ICER'13})}
}
```

vahtras:
Thanks for your comments @gvwilson  and @ntmoore !  As I saw this material used in a workshop by another instructor his reflection was also that the order of the sections was, well odd. Like having an advanced data structure like the dataframe  before the simpler list.

 I can relate to the fact that having an explicit example first  and  general principles later is better for learning. However, that is not what I saw here, my first impression of this material was rather that the order of the sections was somewhat accidental. And I thought reordering sections was a  minor change, not a massive change.

But OK, I respect your reservations and we can leave it as is for now.

PR #299, opened 2018-04-13
Add note about Python 2 integer division

mbonsma:
Since some people may still be using Python 2, and this particular weirdness caused me a lot of grief early in my coding career.


mbonsma:
This is also mentioned in the `Division Types` exercise, but it might still be worth including in the main text. I'm open to other opinions though.

PR #302, opened 2018-04-15
removed for loop and lists from plotting section

timothylwarren:
This is a minor contribution reordering some bits of the plotting episode so as not to mention for loops or lists, which are in a subsequent episode.

### DIFF
--- a/_episodes/09-plotting.md
+++ b/_episodes/09-plotting.md
@@ -75,19 +75,14 @@ plt.ylabel('GDP per capita')
 ~~~
 {: .python}
 ![GDP barplot for Australia](../fig/9_gdp_bar.png)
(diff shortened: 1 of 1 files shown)


souravsingh:
I believe preserving the exercise as a homework would be good idea.

vahtras:
This is an improvement. Yes an exercise later is ok.

alee:
:+1: to moving the accumulator exercises to episode 12. We can add commits to this PR to preserve context or we could make a separate issue for it - thoughts?

vahtras:
Better to uncouple to separate issues in my opinion.

alee:
Done, added #304 to track this

PR #303, opened 2018-04-16
simplify exercise on parson problem and add solution, fixes #288

### LINKED ISSUE #288: Problem too challenging for novices
I taught using this curriculum over the weekend and students found the 
[Jigsaw Puzzle (Parson's Problem) Programming Example](https://github.com/swcarpentry/python-novice-gapminder/blob/gh-pages/_episodes/06-libraries.md#jigsaw-puzzle-parsons-problem-programming-example) in the Libraries episode insurmountably difficult. Few out of 29 solved it without extensive help, we spent a long time on it, many complained at the break, and at least one mentioned it in the post workshop survey.  

> Rearrange the following statements so that a random DNA base is printed. Not all statements may be needed. Feel free to use/add intermediate variables.
> 
> ```python
> bases="ACTTGCTTGAC"
> import math
> import random
> len(bases)
> len(bases)+1
> math.floor(s1)
> math.ceil(s1)
> print("random base ",bases[])
> random.random()*l
> ```

The problem asks students to 
 1. reason through selecting a random element from a list 
 1. learn about the standard library random that is not introduced in the lesson
 1. convert a random number in [0.1) to [0,len-1] - very novel idea for non-programmers
 
To solve it as written they need to:
 - modify lines (implicit fill in the blank)
 - remove lines
 - reorder lines
This is a lot of different types of decisions to make. One of the helpers actually had some learners just delete all of the code that was provided and start from scratch reasoning through the problem. 

I think this exercise should be moved after the other exercises in the lesson, since it's a less direct application of the material in the lesson.  

Possible minor revisions: 
 - make explicit the fill in the blanks [i think this is bare minimum]
 - reduce to reorder and fill in the blank only [this would focus the reasoning on the flow of solving the problem] **this is my vote among minor revisions**
 - reorder and remove only, no fill in the blanks [this would focus the reasoning on ceil vs floor and len vs len +1]
 - reorder only [ this is the easiest]

A related revision is to add a Parson's problem that's just a reorder these lines to an earlier episode to make this type of problem more familiar before doing one with this many required modifications.  

A more significant revision which I would prefer is be to introduce numpy in the lesson and use numpy.random.rand_int(). In my post-workshop debrief, it was brought up that the reason to stick with standard libraries is to not introduce the new concept of importing a submodule. I think the complexity of importing a submodule would follow more directly from the lesson as is than converting a random number in [0.1) to [0,len-1].  Many students asked why it had to be so complicated to just get a random integer and the honest answer is that it doesn't have to be - "we didn't want to introduce a new concept" isn't a good answer to give someone who is stuck and frustrated because the problem as is introduces a new concept that the learners didn't see the utility of(converting random numbers). 

I am happy to make the changes and submit a PR, but I think it's worth some discussion first.

brownsarahm:
I opted to reduce the complexity of the task to to reorder and fill in the blank only to focus the reasoning on the flow of solving the problem, instead of having also delete lines.  I also made fill in

### DIFF
--- a/_episodes/06-libraries.md
+++ b/_episodes/06-libraries.md
@@ -206,22 +206,36 @@ cos(pi) is -1.0
 
 
 > ## Jigsaw Puzzle (Parson's Problem) Programming Example
-> 
+>
 > Rearrange the following statements so that a random
-> DNA base is printed.  Not all statements may be needed.  Feel free to use/add 
-> intermediate variables. 
+> DNA base is printed.  Not all statements may be needed.  Feel free to use/add
+> intermediate variables.
 >
 > ~~~
 > bases="ACTTGCTTGAC"
 > import math
 > import random
-> len(bases)
-> len(bases)+1
-> math.floor(s1)
-> math.ceil(s1)
+> ___ = math.floor(s1)
+> ___ = len(bases)
 > print("random base ",bases[])
-> random.random()*l
+> ___ = random.random()*l
 > ~~~
+> {: .python}
+>
+> > ## Solution
+> >
+> > ~~~
+> > import math
+> > import random
+> > bases="ACTTGCTTGAC"
+> > l = len(bases)
+> > s1 = random.random()*l
+> > idx = math.floor(s1)
+> > print("random base ",bases[idx])
+> > ~~~
+> > {: .python}
+> >
+> {: .solution}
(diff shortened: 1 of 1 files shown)


souravsingh:
Merging. Thanks for the PR @brownsarahm

PR #308, opened 2018-04-30
Made parson's problem less confusing

mikemhenry:
This solution doesn't introduce new functions to students and also
solves the problem in a more straightforward way.


souravsingh:
Looks good to me!

Issue #314, opened 2018-05-09
Remove .python class on "Markdown does most of what HTML does" section


rgaiacs:
On https://raw.githubusercontent.com/swcarpentry/python-novice-gapminder/gh-pages/_episodes/01-run-quit.md, the `.python` class is not necessary.

````
<div class="row">
  <div class="col-md-6" markdown="1">
~~~
*   Use asterisks
*   to create
*   bullet lists.
~~~
{: .python}
  </div>
  <div class="col-md-6" markdown="1">
*   Use asterisks
*   to create
*   bullet lists.
  </div>
</div>
````

should be

````
<div class="row">
  <div class="col-md-6" markdown="1">
~~~
*   Use asterisks
*   to create
*   bullet lists.
~~~
  </div>
  <div class="col-md-6" markdown="1">
*   Use asterisks
*   to create
*   bullet lists.
  </div>
</div>
````

And the same change need to be done for other examples.

PR #316, opened 2018-05-10
Remove entries with {: .python}

vahtras:
as recommended in issue #314

### DIFF
--- a/_episodes/01-run-quit.md
+++ b/_episodes/01-run-quit.md
@@ -38,7 +38,6 @@ keypoints:
     ~~~
     $ jupyter notebook
     ~~~
-    {: .python}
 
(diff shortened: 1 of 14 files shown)


alee:
Hey @vahtras! We should only remove .python from the Markdown examples in episode 01, since those snippets are not Python code. @rgaiacs suggestion in #314 is only referring to the first episode, not the subsequent ones (unless they also have non-Python code snippets marked as `.python`.

Code snippets from the later episodes should instead be updated to `.language-python` as indicated in #312 

Given the size of this change, I think it's probably easier to just open a new PR than do surgery on this PR to remove the changes to all the other episodes. Closing for now.

PR #317, opened 2018-05-15
Array slicing - out-of-range upper index

ysbecca:
An additional challenge question related to array slicing which explores what happens when an upper index out-of-range is used. It is helpful to know that no error is thrown, but behaviour may not be as expected.


ysbecca:
Thanks for the feedback. I did notice there are some missing details, I have submitted another pull request with brief proposed solutions, if this is helpful.

vahtras:
OK, I am merging this looking forward to an update on the solution part.

Issue #311, opened 2018-05-03
Parson's Problem exercise typo


colinsauze:
There's a line in the code to be used in parson's problem exercise in the libraries section which says:
print("random base ", bases[])
the solution is 
print("random base ", bases[idx])
A better text code the quiz would either have the variable name between the [  and ] or have ___ to indicate that the student needs to work out which to put here themselves. Or was there some intention to cause an error for them to solve?

PR #319, opened 2018-05-22
Small Modification to Example in Lesson 6

martintb:
Issue #311 points out that one of the examples in this lesson uses an
inconsistent style with other examples. The necessary change is to add
underscores in the example to indicate where students should add code.


vahtras:
Thanks!

Issue #175, opened 2017-06-09
Jupyter only accept one language per notebook


rgaiacs:
http://swcarpentry.github.io/python-novice-gapminder/01-run-quit/ says

> this format allows Jupyter to mix software (in several languages) with documentation and graphics, all in one file.

But the statement isn't 100% true because on Jupyter Notebook the kernel/language is associate with the notebook itself and not with the cell.

Can we change the statement to

>  this format allows Jupyter to mix source code, text and images, all in one file.

mzc9:
May I take a stab at this?

alee:
Absolutely! 👍

PR #324, opened 2018-06-02
Verbiage re Jupyter Notebooks Issue #175

mzc9:
Changed verbiage in http://swcarpentry.github.io/python-novice-gapminder/01-run-quit/


alee:
Thanks!

Issue #325, opened 2018-06-04
Add instructions on saving figures to "Plotting" lesson


gaurav:
The [*Plotting* lesson](https://swcarpentry.github.io/python-novice-gapminder/09-plotting/) currently ends by creating various plots using `matplotlib.pyplot`. I think this lesson should end with instructions on saving the figures to disk as images using [`matplotlib.pyplot.savefig`](https://matplotlib.org/api/_as_gen/matplotlib.pyplot.savefig.html). This would allow students to see how Jupyter Notebook can help them get from input data to images they could use in their papers and publications, and provides a sense of completion at the end of this set of data ingest/plotting lessons before turning back to the subsequent lessons learning about how to use [Python lists](https://swcarpentry.github.io/python-novice-gapminder/11-lists/), [for loops](https://swcarpentry.github.io/python-novice-gapminder/12-for-loops/) and so on.

souravsingh:
+1 on the suggestion.

vahtras:
+1

souravsingh:
@gaurav Would you be willing to create a PR for the issue?

gaurav:
@souravsingh I'd rather not -- I'm not very familiar with PyPlot, so I'm not sure if I can provide the best or most idiomatic way of exporting from it, and I don't know if there are any Jupyter tricks I should use in the export. Besides, this does make a good first issue for new contributors! However, if nobody's done anything about this in a month or so, let me know and I'll give it a stab?

souravsingh:
Fixed in #331. Closing

PR #331, opened 2018-06-21
include savefig

vahtras:
Proposition that adresses issue #325, suggested by @gaurav. Is this what you had in mind? Feel free to comment, update. I made it as a callout box in the end.

Regards,
Olav


gaurav:
Yup, that's exactly what I had in mind! Thanks so much @vahtras!

PR #336, opened 2018-08-01
08-data-frames.md update, apply and aggregate descriptions

sarahymurphy:
I added a bit of description below the section about using the apply function and the aggregate function. I think these can be confusing for learners as they're a bit higher level than the information leading up to it. I've also had a bit of trouble teaching this part because I find it difficult to describe. 

I'm not sure if I worded my description in the most eloquent way, but I feel that this part of the lesson needs at least a bit of a description to help both instructors and learners using this episode.

### DIFF
--- a/_episodes/08-data-frames.md
+++ b/_episodes/08-data-frames.md
@@ -260,6 +260,8 @@ mask_higher = data.apply(lambda x:x>x.mean())
 wealth_score = mask_higher.aggregate('sum',axis=1)/len(data.columns)
 wealth_score
 ~~~
+{: .python}
+
 By using lambda, we can create one-line functions to apply to dataframes. This code block applies a
 lambda function to x. The format used is `lambda variable: function` in which the function uses 
 the variable defined on the left side of the colon. For each value of x, we are applying the conditional
@@ -268,7 +270,7 @@ checking if x is greater than the mean of all values of x.
 Aggregate applies operations to the dataframe. Here, we are using aggregate to apply the operation 'sum'
 over each column (axis = 1) in the dataframe.
 
-{: .python}
+
 ~~~
 country
 Albania                   0.000000
--- a/_episodes/08-data-frames.md
+++ b/_episodes/08-data-frames.md
@@ -267,7 +267,7 @@ checking if x is greater than the mean of all values of x.
 
 Aggregate applies operations to the dataframe. Here, we are using aggregate to apply the operation 'sum'
 over each column (axis = 1) in the dataframe.
-~~~
+
 {: .python}
 ~~~
 country
--- a/_episodes/08-data-frames.md
+++ b/_episodes/08-data-frames.md
@@ -260,6 +260,14 @@ mask_higher = data.apply(lambda x:x>x.mean())
 wealth_score = mask_higher.aggregate('sum',axis=1)/len(data.columns)
 wealth_score
 ~~~
+By using lambda, we can create one-line functions to apply to dataframes.. This code block applies a
+lambda function to x. The format used is `lambda variable: function` in which the function uses 
+the variable defined on the left side of the colon. For each value of x, we are applying the conditional
+checking if x is greater than the mean of all values of x. 
+
+Aggregate applies operations to the dataframe. Here, we are using aggregate to apply the operation 'sum'
+over each column (axis = 1) in the dataframe.
+~~~
 {: .python}
 ~~~
 country
--- a/_episodes/08-data-frames.md
+++ b/_episodes/08-data-frames.md
@@ -260,7 +260,7 @@ mask_higher = data.apply(lambda x:x>x.mean())
 wealth_score = mask_higher.aggregate('sum',axis=1)/len(data.columns)
 wealth_score
 ~~~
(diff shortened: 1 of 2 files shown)


souravsingh:
Looks good to me. Anybody else want to take a look?

alee:
Thanks for the PR - this is a great start! I'd like to refine the description of lambda and how it interacts with a dataframe via apply though. What do you think about adding links to the [python lambda tutorial](https://docs.python.org/3.6/tutorial/controlflow.html#lambda-expressions) and the [pandas dataframe docs](https://pandas.pydata.org/pandas-docs/version/0.23.3/generated/pandas.DataFrame.apply.html) also?

A minor pedantic note is that we should probably follow pandas documentation and example guidelines and name the `data` variable `df` or `data_frame`..

I'll take a closer look at it this evening with some more concrete suggestions unless another maintainer wants to chime in first!

sarahymurphy:
Thanks for the feedback! I like the idea of including those links - that documentation is super helpful when trying to wrap your head around the lambda stuff.

alee:
Sorry for the long delay @sarahymurphy ! The evening I had available to take a closer look got swept away by monsoon storms and a multitude of other lame excuses.

I think I'd like to reframe the paragraph a bit and maybe expand it out so the various commands are more sequential and the instructor takes the time to inspect the output of each command being entered. The lambda function is an anonymous function that will be applied to each column of the DataFrame when axis=0, which is the default or each row when `axis=1`. In this particular case we don't specify the axis in our `data.apply()` call, and so the default `axis=0` applies and thus the `lambda x: x > x.mean()` gets applied to each column in the DataFrame, asking it "Hey, for each row in this column, is the given value greater than the mean of all values in the column? This results in the following DataFrame:

```
gdpPercap_1952	gdpPercap_1957	gdpPercap_1962	gdpPercap_1967	gdpPercap_1972	gdpPercap_1977	gdpPercap_1982	gdpPercap_1987	gdpPercap_1992	gdpPercap_1997	gdpPercap_2002	gdpPercap_2007
country												
Albania	False	False	False	False	False	False	False	False	False	False	False	False
Austria	True	True	True	True	True	True	True	True	True	True	True	True
Belgium	True	True	True	True	True	True	True	True	True	True	True	True
Bosnia and Herzegovina	False	False	False	False	False	False	False	False	False	False	False	False
Bulgaria	False	False	False	False	False	False	False	False	False	False	False	False
Croatia	False	False	False	False	False	False	False	False	False	False	False	False
Czech Republic	True	True	True	True	True	True	False	False	False	False	False	False
Denmark	True	True	True	True	True	True	True	True	True	True	True	True
Finland	True	True	True	True	True	True	True	True	True	True	True	True
France	True	True	True	True	True	True	True	True	True	True	True	True
Germany	True	True	True	True	True	True	True	True	True	True	True	True
Greece	False	False	False	False	True	False	False	False	True	False	True	True
Hungary	False	False	False	False	False	False	False	False	False	False	False	False
Iceland	True	True	True	True	True	True	True	True	True	True	True	True
Ireland	False	False	False	False	False	False	False	False	True	True	True	True
Italy	False	False	False	False	False	False	True	True	True	True	True	True
Montenegro	False	False	False	False	False	False	False	False	False	False	False	False
Netherlands	True	True	True	True	True	True	True	True	True	True	True	True
Norway	True	True	True	True	True	True	True	True	True	True	True	True
Poland	False	False	False	False	False	False	False	False	False	False	False	False
Portugal	False	False	False	False	False	False	False	False	False	False	False	False
Romania	False	False	False	False	False	False	False	False	False	False	False	False
Serbia	False	False	False	False	False	False	False	False	False	False	False	False
Slovak Republic	False	False	False	False	False	False	False	False	False	False	False	False
Slovenia	False	False	False	False	False	True	True	True	False	False	False	True
Spain	False	False	False	False	False	False	False	False	True	True	True	True
Sweden	True	True	True	True	True	True	True	True	True	True	True	True
Switzerland	True	True	True	True	True	True	True	True	True	True	True	True
Turkey	False	False	False	False	False	False	False	False	False	False	False	False
United Kingdom	True	True	True	True	True	True	True	True	True	True	True	True
```

Next, we aggregate across all rows and sum up the occurrence of `True` values in each row, to get something like this:
```
country
Albania                    0
Austria                   12
Belgium                   12
Bosnia and Herzegovina     0
Bulgaria                   0
Croatia                    0
Czech Republic             6
Denmark                   12
Finland                   12
France                    12
Germany                   12
Greece                     4
Hungary                    0
Iceland                   12
Ireland                    4
Italy                      6
Montenegro                 0
Netherlands               12
Norway                    12
Poland                     0
Portugal                   0
Romania                    0
Serbia                     0
Slovak Republic            0
Slovenia                   4
Spain                      4
Sweden                    12
Switzerland               12
Turkey                     0
United Kingdom            12
```

Finally, we divide each of these values by the total number of gdp measurements taken to derive a measure between 0 and 1 representing how often a given country's GDP has been greater than the mean.

I don't know if we should include these full tables in the lesson but I think it'd be worthwhile to break those steps up a bit more so learners get a better understanding of the intermediate steps being taken and what data transformations are happening and why.

Whew!

Issue #277, opened 2018-03-07
misleading answer to exercise "Exploring the Math Module"


wikfeldt:
In the answer to the "Exploring the Math Module" exercise in episode 6, where `math.pow()` and `math.sqrt()` are compared, it says:   
`... However, sqrt(x) doesn’t support negative arguments.`

`math.pow()` doesn't support negative arguments either (i.e. `math.pow(-1,0.5)`) gives a `ValueError`. So instead perhaps refer to the `cmath` library

souravsingh:
'math.pow(-1, 2)' works without problem.

wikfeldt:
yes absolutely, so the answer is not incorrect. Just a bit misleading (although maybe it's just me) because the discussion is about comparing sqrt and pow for taking square roots and the specific example `pow(x,0.5)` is brought up in the same context as negative arguments

lo5an:
I just opened a PR removing the misleading sentence.  I believe I added this answer, and I was probably confusing `pow()` and `math.pow()` when I wrote it.

PR #338, opened 2018-08-20
Remove potentially misleading note on math.pow

lo5an:
Removes a potentially misleading sentence from the exercise "Exploring the Math Module" in Episode 6, making the answer more correct, and addressing issue #277.


lo5an:
Sure! I'll add some changes to the PR.

PR #340, opened 2018-09-11
Clarified objectives and variable definition

dloney:
The objectives and variable/value definition were difficult to follow for first time learners. Scalar values were simplified to numbers, and data replaced with information. The concept of information was then related to specific pieces which are assigned as values. This is intended to better connect variables with specific pieces of information assigned that are assigned as values.


vahtras:

PR #343, opened 2018-11-08
Fixed solution of "Simulating a dynamical system"

SBolo:
The function iterate is not working, because of an IndexError. I propose the following solution using append.


alee:
Looks good to me as well! I really like the visualization you added @ntmoore but perhaps we need to unpack that code more?

PR #344, opened 2018-11-09
add bifurcation diagram to the logistic map code

ntmoore:
This code makes a nice picture from the logistic map's recursion relation.  The code to produce the figure is fairly long when compared to most homework solutions. I used inline comments - I don't claim this is the best way to present the solution.

@Autoplectic @SBolo @davidrpugh @alee

Here's the image the code should produce 
image

### DIFF
--- a/_episodes/14-writing-functions.md
+++ b/_episodes/14-writing-functions.md
@@ -545,25 +545,39 @@ result of call is: None
 > > # we'll use a few libraries
 > > import matplotlib.pyplot as plt
 > > %matplotlib inline
-> > import numpy as np
 > > import random
 > > # using lists to store the "final" values in the recursion series
-> > r_list=[]
-> > final_value=[]
+> > r_list = []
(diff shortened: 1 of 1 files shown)


Autoplectic:
it might make more sense to do:

    r_list = np.arange(2.4, 4.0, 0.0001)

and then:

    for r in r_list:

as it is, you create an array of r values, and then create a list which is a copy of it.

davidrpugh:
If you really want to avoid using NumPy `arange` function, then you could have another exercise to write a function create a list of floats.  One such solution would be...

```python
def range_of_floats(start, stop, step):
    rs = []
    while start < stop:
        rs.append(start)
        start += step
    return rs

# call this within the the bifurcation plotting code
rs = range_of_floats(2.4, 4.0, 0.003)
````

...could then mention that NumPy already provides this functionality and then link to docs for `arange`.

vahtras:

PR #345, opened 2018-11-13
include maintainers in this citation?

ntmoore:
I believe the citation should include the maintainers.  Is the bi-annual DOI for each lesson still a feature?  This is probably something we should endeavor to get done by June.  Specifically, https://docs.carpentries.org/topic_folders/lesson_development/lesson_release.html


vahtras:
I agree!

souravsingh:
No problem from my side

PR #347, opened 2018-11-15
episode 02 cleanup

deppen8:
- [x] update code chunks to use `.language-python`
- [x] bold first instance of
    - [x] variables
    - [x] arguments
    - [x] index
    - [x] substring
- [x] fix unneeded nested lists
- [x] italicize instead of bold emphasized terms
- [x] 'create 2 cells' --> 'create two cells'
- [x] add inline code markup to a variety of items like `42` and `'AB'`
- [x] split Slicing into two challenges
- [x] add a solution to the new Slicing challenge


souravsingh:
Merging

PR #348, opened 2018-11-18
removed numpy reference

ntmoore:
Per suggestions, I revised the problem to NOT use the numpy arange() function.  This is a little more clunky, but it does work and fits within the introduced tools


vahtras:
Agreed.
Pep8-recommendations should followed as much as possible but not blindly.
Some editors warn about non-compliance or one can use tools such as pylint
to analyze the source. Other examples are, spacing after commas, around
operators.
-Olav

PR #352, opened 2018-11-28
variable data renamed to df in 08-data-frames.md

heijer:
variable renamed to be consistent with previous exercise


vahtras:
Thanks!

PR #353, opened 2018-11-28
wording change in 08-data-frames.md

heijer:
column is the second axis (not the second column)


vahtras:
Thanks! I agree that a reformulation is in place, but I do not agree with your suggestion; axis is a parameter that has values 0 or 1 to signal that we are talking about rows or columns. Looking at the example,  'continent' is actually the first column and I belive the word index should be left out in this context, since the index has a special meaning in data frames (here  the 'country' column is set to index by the read_csv call).

vahtras:
Closing this now due to transitioning to Carpentries Workbench. You are welcome with a new PR after the transition.

PR #358, opened 2018-12-17
Slightly restructure episode 12

xorJane:
Hello!

Explaining what a `for` loop is made of (in a language agnostic way) seems more important than Python's exact syntax for a `for` loop. This PR restructure episode 12 slightly so that the key point "A `for` loop is made up of a collection, a loop variable, and a body." comes before "The first line of the `for` loop must end with a colon, and the body must be indented."

If you agree with the intent of this PR, do you think "The first line of the `for` loop must end with a colon, and the body must be indented." should come even after the individual sections on variables, ranges, and loop bodies? Or is it best positioned where I've moved it for now?

Thanks for reviewing!

### DIFF
--- a/_episodes/12-for-loops.md
+++ b/_episodes/12-for-loops.md
@@ -10,9 +10,9 @@ objectives:
 - "Write for loops that use the Accumulator pattern to aggregate values."
 keypoints:
 - "A *for loop* executes commands once for each value in a collection."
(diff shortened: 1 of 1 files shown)


xorJane:
Thanks for the feedback!

xorJane:
Thanks for merging!

PR #371, opened 2019-02-16
Average program should have a divide by count

anzelpwj:
The `average` program was calling itself, and should have sum/len.


vahtras:
thanks!

PR #373, opened 2019-02-21
Moved discussion of cell types to introduction of cells

### LINKED ISSUE #332: Clarify "code vs. text" and "command vs. edit" discussion
# Issue
In `Running and Quitting`, under the section `The Notebook has Command and Edit modes`, there is a pinned note and paragraph titled `Code vs. Text`. This content seems out-of-place for the main topic of commanding vs. editing that it is under.  There is no particular relevance to the list of command shortcuts that do belong to the section.

# Suggestion
Move the content in question to just before `The Notebook will turn Markdown ...`, possibly giving it a bit more time with a discussion of cell types.

pnb:
This change moves the discussion of code vs. text cell types up to the first mention of cell types. I also added a very brief transitional sentence to introduce the word cell, since cells were previously referred to as blocks in the lesson.

This fixes #332, although I took a slightly different approach from the suggestion in that issue and moved the discussion up instead of down. I thought that it made the most sense to discuss cell types when the different types are first introduced.

### DIFF
--- a/_episodes/01-run-quit.md
+++ b/_episodes/01-run-quit.md
@@ -63,19 +63,18 @@ keypoints:
 > *   But this format allows Jupyter to mix source code, text, and images, all in one file.
 {: .callout}
 
-## The Notebook has Command and Edit modes.
-
-*   Open a new notebook from the dropdown menu (that says 'New') in the top right corner of the file browser page.
-*   Each notebook contains one or more cells that contain code, text, or images.
-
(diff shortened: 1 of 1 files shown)


vahtras:
Thanks, looks good to me!

alee:
👍 from me as well, thanks for the PR!

souravsingh:
Merging.

PR #380, opened 2019-04-05
Change 'array' to 'list

lexnederbragt:


lexnederbragt:
Just found another small mistake with the exercise which this PR now also fixes...

PR #385, opened 2019-04-11
Clearer * works on strings not only single chars

steltenpower:
The diff is small and speaks for itself really


vahtras:
I agree with @alee on this.

martinosorb:
Not a maintainer but I agree with @alee as well.

vahtras:
Closing for the transition to Carpentries Workbench. I believe the consensus is to not include the suggested change. We do welcome new PRs after the transition has been completed.

Issue #382, opened 2019-04-05
Pandas lessons - Using pandas alias


cschlosberg:
In the Pandas DataFrames and Reading Tabular Data into DataFrames lessons, I think pandas should be imported as an alias:
```import pandas as pd``` instead of ```import pandas```. Then each of the function can be used aliased: ```data = pd.read_csv('data/gapminder_gdp_oceania.csv')``` instead of ```data = pandas.read_csv('data/gapminder_gdp_oceania.csv')```. I find this to be very common practice and is supported in the [pandas documentation](http://pandas.pydata.org/pandas-docs/stable/getting_started/10min.html). In addition, it can be create consistency across lessons. For example, in the Plotting lesson, matplotlib is imported as an alias: `import matplotlib.pyplot as plt`.

alee:
I agree, this would be a good change for consistency since it is pretty established convention (across official documentation, examples, and existing code) to `import pandas as pd` and `import numpy as np`

rantahar:
I added pull requests for the change. I think it's a good idea to mention this pattern, since it is common in examples and discussion, as @cschlosberg says, and often goes unexplained. This will also reinforce aliasing, which is introduced in the lesson 5.

I'm not sure if the alias should be used in the text, as opposed to code blocks. I chose to change any references to pandas.function to pd.function also in the text for consistency.

PR #391, opened 2019-04-12
Import Pandas as pd in lesson 17, conditionals

rantahar:
Import Pandas using the common alias 'pd' and use the alias in the examples and exercises.

Related to issue #382. Aliasing itself is introduced in the previous lesson and using it here would reiterate the topic. The shorthand itself is common and often assumed in examples and discussion, so I think it would be useful to introduce.


alee:
thanks for all the PRs!

PR #396, opened 2019-05-13
Add mention of the file browser opened by jupyter notebook

### LINKED ISSUE #395: Learners confused about creating new notebooks
I have a coworker who was walking through these lessons to get an idea of python. She was confused pretty early on in "Running and Quitting" because it tells the reader how to start the jupyter server and then shows a screenshot of an open notebook. However, the reader hasn't been told how to create a new notebook yet so their screen (browser window showing the jupyter file browser) does not look like the screenshot. The instructions for creating a notebook are after the screenshot.

I'm not sure if this is worth changing since these lessons are typically taught by an instructor, but it is a little confusing for anyone reading the lessons on their own. I thought I'd at least start the discussion.

djhoese:
Close #395.

As mentioned in #395 there is currently no mention of the file browser that is first shown in the browser window when starting the Jupyter Notebook server with `jupyter notebook`. A coworker was confused at why their screen (the file browser) was not the same as the screenshot shown.

I could see a better solution being a screenshot of the file browser before the current screenshot, but I'm not sure what files should be shown in the screenshot or if there are any preferred ways of making screenshots for lessons.

Feel free to ask me to change the language/tone of what I've changed. Thoughts?

### DIFF
--- a/_episodes/01-run-quit.md
+++ b/_episodes/01-run-quit.md
@@ -40,6 +40,8 @@ keypoints:
     ~~~
 
 *   This will start a Jupyter Notebook server and open your default web browser. 
+*   The server first opens a file browser page where new notebooks can be created
+    or existing notebooks can be opened.
 *   The server runs locally on your machine only and does not use an internet connection.
 *   The server sends messages to your browser.
 *   The server does the work and the web browser renders the notebook.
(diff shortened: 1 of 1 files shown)


alee:
Note that the changes here may be subsumed by #398 now - would you like to fit these changes into that larger PR?

alee:
Closing for now as this has been mostly addressed in #398 and the accompanying JupyterLab UI changes.

PR #403, opened 2019-05-21
reference.md: fix code blocks

maxim-belkin:
Use proper syntax for Python code blocks.


maxim-belkin:
> but do we want to keep the `$` prompt for `$ jupyter notebook`?

I believe `$` is used in standalone code blocks (between `~~~` with appropriate code block specifications after the second `~~~`) and I did not see any other uses of `$` in inline code blocks in this lesson. Shall I revert it back?

alee:
I'm fine with it either way, will let the other maintainers chime in to see what they think.. thanks again!

PR #406, opened 2019-05-22
Add binder button

davidrpugh:
This PR adds an `environment.yml` file that is used by Binder to install Python 3.7 and the most recetn version of Anaconda Python and adds a Binder button to the README. Note that for testing purposes I had to use the `add-binder-button` branch on my fork to get Binder to build the image. I will leave it as is for now so that it will work when you click the button but I will need to change the url before PR is merged.

Also, this is the simplest way that could possible work.  Everything in the lesson repo is mapped into the working directory in the Binder instance. The working directory is a bit cluttered and this might not be desirable.  Anyway feedback is most appreciated!


davidrpugh:
Here is the working link to the binder instance...

https://mybinder.org/v2/gh/kaust-vislab/python-novice-gapminder/add-binder-button?urlpath=lab

This will need to be changed to 

https://mybinder.org/v2/gh/swcarpentry/python-novice-gapminder/gh-pages?urlpath=lab

prior to merging.

davidrpugh:
@alee @vahtras Feedback on this PR would be helpful.  

* This is the simplest prototype that works. When you click the binder button, the Jupyterlab interface launches on Binder service. 
* The `environment.yml` file used by Binder/Conda installs only Python, Matplotlib, Jupyterlab, and Pandas (and their deps) instead of the full Anaconda distribution.

I played around with getting the episodes to render properly as separate panels in the Jupyterlab UI so that you could have an episode and a notebook/console open at the same time but could not get the Markdown or even the html to render properly.  I think the entire `_site` directory needs to be served via `jekyll` in order for everything to work as expected and this might require a new juyterlab extension.

davidrpugh:
@alee I agree. When I have used binders in my own workshops this is exactly what I do. I have a second repo that has workshop specific binder buttons in the README. The repo contains the binder configurations files and a data directory containing the gap under data files.

Do you have repo creation rights with swcarpentry GitHub .org?

alee:
Yep, looks like I do - I just made https://github.com/swcarpentry/python-novice-gapminder-binder and added you to the repo. It's still private for now until we decide you are ready for release. Thanks!

davidrpugh:
I added all of he required files.  Now all you need to do is make the repo public and then you can click the button!

davidrpugh:
@alee another option, that would avoid the creation of an additional repo, would be to create an orphaned branch called `binder` which would initially be empty and then we would install those files required to run Binder as well as the data and a README into that branch.  Binder can then build off of that branch. The binder button would still go on the `gh-pages` README which would be easier to discover.  Thoughts?

djinnome:
Hi @alee 
What  @davidrpugh describes is how I incorporated binder into the [cell factory design workshop](https://agilebiofoundry.github.io/2019-05.01-cell-factory-design-brownbag/setup/)
The gh-pages branch setup page contains the binder button and the master branch contains the actual notebooks, data files and environment.yml file that binder uses to install all the prereqs.

The other modification we made is instead of converting the jupyter notebook to markdown to be rendered in gh-pages, we included an iFrame in each gh-pages episode to nbviewer.jupyter.org which renders the notebooks in the master branch. See, for example, [episode 1: Intro to FBA](https://agilebiofoundry.github.io/2019-05.01-cell-factory-design-brownbag/01-Intro-to-FBA/)

alee:
Sounds good to me - I just set up an orphan branch https://github.com/swcarpentry/python-novice-gapminder/tree/binder - do you want to close this PR and submit a new one against that branch? 

Sorry for all the extra work! :hammer_and_pick:

davidrpugh:
@alee I have put all of the necessary files in the `binder` branch to generate a Binder instance and then removed the environment file from this PR. The only change this PR should now be making is to add the Binder button.

alee:
Thanks @davidrpugh ! Would you mind opening a new PR though? We would have a cleaner commit history if we only include the necessary changes which are fairly minimal now due to the binder branch.

davidrpugh:
Sure.

PR #417, opened 2019-06-05
Use display() instead of print() to render html-formatted DataFrames

djinnome:
Hi folks,

 I went ahead and opened a pull request for episode 7 based on feedback from issue [#342](https://github.com/swcarpentry/python-novice-gapminder/issues/342) switching `print()` to `display()`

Let me know if it displays correctly in gh-pages, and I'll go ahead and generate a corresponding pull request for episode 8.

### DIFF
--- a/_episodes/07-reading-tabular.md
+++ b/_episodes/07-reading-tabular.md
@@ -22,11 +22,11 @@ keypoints:
 *   Borrows many features from R's dataframes.
     *   A 2-dimensional table whose columns have names
         and potentially have different data types.
-*   Load it with `import pandas as pd`. The alias pd is commonly used for Pandas.
-*   Read a Comma Separate Values (CSV) data file with `pd.read_csv`.
+*   Load it with `import pandas as pd`. The alias `pd` is commonly used for Pandas.
+*   Read a Comma Separated Values (CSV) data file with `pd.read_csv`.
     *   Argument is the name of the file to be read.
     *   Assign result to a variable to store the data that was read.
-* In a jupyter notebook, use `display()` not `print()` to see view the output of a dataframe.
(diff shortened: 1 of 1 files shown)


djinnome:
How does one make individual tables scrollable? Do you need to use an iFrame or something?

djinnome:
Unless you mean using [`qgrid`](https://qgrid.readthedocs.io/en/latest/)?

djinnome:
Are you talking about the changing the behavior of the `display()` function to generate scrollbars for the pandas dataframe output, or are you talking about adding a scroll bar to  the HTML table in the Instructor's markdown document?

This brings up another point. It might be better to actually render a jupyter notebook with an iFrame instead of converting it to Markdown.  This  [Lesson on cell factory design](https://agilebiofoundry.github.io/2019-05.01-cell-factory-design-brownbag/01-Intro-to-FBA/)

alee:
Sorry for the latency, was 🏕 in the Grand Canyon for a few days with the family 😅 

Hmm, I think both, and I was mostly thinking out loud / checking to see if it's possible. I find the new display tables harder to read as they are wide and require side scrolling but that's my personal preference.. other's thoughts?

djinnome:
Hi @alee sounds like a wonderful vacation. It is good to unplug every once in a while.  Which browser are you using?  I am using Chrome and I see side scroll bars for the tables so I don't have to page scroll.

Talha-Naeem:
image

Talha-Naeem:
%matplotlib inline
it works and this command tell by my Sir Anees Ahmed

vahtras:
Closing inactive PRs with unresolved conflicts before transitioning to Carpentries Workbench.

PR #423, opened 2019-06-30
Add necessary files to the binder branch

davidrpugh:
This PR add necessary files to the `binder` branch of the `python-novice-gapminder` lessons. Will open a separate PR to add the Binder button to the main `README.md`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
 [![Create a Slack Account with us](https://img.shields.io/badge/Create_Slack_Account-The_Carpentries-071159.svg)](https://swc-slack-invite.herokuapp.com/) 
- [![Slack Status](https://img.shields.io/badge/Slack_Channel-swc--py--gapminder-E01563.svg)](https://swcarpentry.slack.com/messages/C9X4W03KL) 
+[![Slack Status](https://img.shields.io/badge/Slack_Channel-swc--py--gapminder-E01563.svg)](https://swcarpentry.slack.com/messages/C9X4W03KL) 
 
 python-novice-gapminder
 =======================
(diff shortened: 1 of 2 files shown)


davidrpugh:
@alee Do we need to include the `_site` directory in this repo? I thought that we had agreed that it would be best not to include the lesson materials inside the Binder instance.

davidrpugh:
@alee Let's finalize what we want to include in this branch. Once this PR is merged, then you can merge PR #406 which adds the binder button to the `gh-pages/README.md`.

djinnome:
I have a few thoughts on this topic.  There are a number of arguments for including or not including different aspects of the lesson materials, depending on the use case. Let's enumerate the possibilites:

1. **Live workshop student**: completely empty binder with only the data and prerequisites, and no notebooks.
2. **Teacher practice**: skeleton notebooks, with markdown, but empty code cells.
3. **Self-study**:  notebooks with markdown and code cells, but no outputs
4. **Teacher solution manual**: notebooks with markdown and code cells, with outputs.

My thought is that each of these could be a different branch, with a different binder URL, corresponding to the different use cases.

davidrpugh:
@djinnome the use case that I had in mind was your number 1 (so I removed the `notebooks` dir to have clean repo with only the `data` director and the `environment.yml` file.  I think use cases 2-4 make sense as separate branches but I think we should make those separate PR(s).  

@alee apologies but for some reason I thought the `_site` directory was included in the branch because it was showing up my my local machine but it is being ignored.

davidrpugh:
@alee JupyterLab 1.0 is out on `conda-forge` so I added the channel and bumped the version number, once Jupyterlab 1.0 is released on the `main` channel then I will remove the `conda-forge` channel to keep the `environment.yml` file as simple as possible.

The environment file will already download Pandas 0.24.2: if you specify the major and minor version numbers but leave the patch version unspecified, then Conda will download the most recent patch by default.

alee:
Thanks for the clarification! I didn't realize patch versions would be included by default, I don't use conda much, typically just pipenv or pip :-).

Looks good to me, I'll merge shortly.

PR #424, opened 2019-07-01
Removed unnecessary files

davidrpugh:
This PR removes two files which I don't think are necessary for the `binder` branch.


davidrpugh:
@alee I think we can remove these two files from the `binder` branch they don't seem to be necessary and it would be good to keep this branch as close as possible to an ideal project setup for Python projects.

alee:
Thanks!

PR #428, opened 2019-07-16
Refactor binder button

davidrpugh:
This PR customizes the binder button so that it says "Launch JupyterLab" instead of "Launch Binder". I think this change will make it more obvious for users what is going on when they click the button.


davidrpugh:
Sure...once I figure out exactly how to squash commits! Yep I do have an AMY profile and you can find me on the instructors page...

https://carpentries.org/instructors/#davidrpugh

davidrpugh:
I am just going to start over with this PR...

Issue #111, opened 2016-10-05
More "scientific" examples in the for loops lesson


leouieda:
The instructor training [checkout instructions](http://swcarpentry.github.io/instructor-training/checkout/) specify:

> Try to give exercises a scientific theme: reversing a string isn’t particularly motivational for our audience.

I noticed that is exactly the topic of one of the challenges in the "For loops" chapter. Is there a purpose for this? If not, I could try to come up with a better one (like reversing a list of file names, dates, or something similar).

gvwilson:
Sure - the challenge is to only use things that we've already introduced.

eldobbins:
I was looking for a possible issue to solve for my first contribution to this lesson.  I came across this because of the labels.  However, I'm unclear if this is an issue with this exercise specifically, or a general comment about how to design exercises.

If it's a problem with just this exercise, I'd like to try and fix it.

If it's a general comment, perhaps the labels should be changed.

PR #431, opened 2019-07-19
updated loop example

isaac-galvan:
Hello,

I added a different example based on the feedback from #111 

Thanks for reviewing!

### DIFF
--- a/_episodes/12-for-loops.md
+++ b/_episodes/12-for-loops.md
@@ -211,25 +211,24 @@ print(total)
 > {: .solution}
 {: .challenge}
 
-> ## Reversing a String
+> ## Looping a Sliced List
 >
-> Fill in the blanks in the program below so that it prints "nit"
-> (the reverse of the original character string "tin").
+> Fill in the blanks in the program below so that it prints every other list item after skipping the first item.
 >
 > ~~~
-> original = "tin"
+> original = ['Ireland', 'Italy', 'Montenegro', 'Netherlands']
 > result = ____
-> for char in original:
->     result = ____
+> for country in original[____]:
+>     result._____(country)
 > print(result)
 > ~~~
 > {: .language-python}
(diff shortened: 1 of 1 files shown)


souravsingh:
The PR looks good to me.

souravsingh:
I can merge this if no one else has any problems

alee:
Thanks for the PR @isaac-galvan ! However, I'm not quite convinced that this a clear improvement over the original example. This for loop could also be boiled down to `results = original[1::2]` which raises the question of why even use a for loop.

Perhaps we can lift one of the examples from http://swcarpentry.github.io/python-novice-inflammation/02-loop/index.html - IMO it makes sense for the @swcarpentry/python-novice-gapminder-maintainers and @swcarpentry/python-novice-inflammation-maintainers to develop a shared set of compelling examples since there's a lot of overlap between the lessons.

maxim-belkin:
Thanks for tagging us, @alee.

With regards to "reversing a string" exercise: I tend to agree that it has to be changed because (a) how often do we loop over a string in real life? and (b) we can achieve the same effect with `[::-1]`. I also agree with @alee's concerns that suggested exercise can be "solved" with `[1::2]` so... it's not that much different. It seems that we need an exercise that is a tiny bit more complex. Complexity can come from the need to check certain quality(-ies) of the list elements -- first character, type, etc.

Our lesson talks about loops before lists so we can't actually do this switch right now: there is an effort to change the order of episodes but it's a WIP. Your lesson talks about lists before loops, so you're all good.

eldobbins:
Update: it seems the order of the lessons has been changed so that now lists comes before loops. So we should be able to make progress now on the PR.  But if this solution has been rejected, perhaps this PR should be closed?

If not closed, is it possible for me to modify it in my fork?  Not sure of the git procedure here.

PR #436, opened 2019-08-21
Restructure lesson 9

ChrisKSeal:
I would like to propose restructuring lesson 9: Plotting. The lesson begins with a simple plot using two lists and matplotlib. This is followed by using the inbuilt plot function with a dataframe, before extracting data from the aforementioned dataframe and using this with matplotlib.
I propose that we move the use of the inbuilt plot function to after the point in which the data is extracted from the data frame and plotted with matplotlib. In my opinion this provides the learners with a linear progression from list -> column in dataframe -> function within dataframe that clarifies the different ways data can be incorporated into a plot.

### DIFF
--- a/_episodes/09-plotting.md
+++ b/_episodes/09-plotting.md
@@ -39,11 +39,9 @@ plt.ylabel('Position (km)')
 {: .language-python}
 
 ![Simple Position-Time Plot](../fig/9_simple_position_time_plot.svg)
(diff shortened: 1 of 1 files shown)


Accio:
I agree with this suggestion. As I go through lesson 9, assuming I have no background knowledge, I am lost.

vahtras:
Closing inactive PRs with unresolved conflicts before the transition to Carpentries Workbench. I do think it is a good suggestion but it cannot be incorporated at this point. Welcome with an updated PR after the transition has been completed.

PR #440, opened 2019-09-04
merge styles for UI fixes and updated docker jekyll

alee:


alee:
closing for now, waiting for upstream styles changes to the Docker targets that use bundler https://github.com/carpentries/styles/pull/428/files

PR #443, opened 2019-10-01
Show popular StackOverflow questions, not newest

katrinleinweber:
Is it likely that our learners want to see whichever questions were asked there recently? If not, please consider this update that sends them to the popular questions (which IMHO are likely more relevant and of exemplary quality).

### DIFF
--- a/_episodes/19-wrap.md
+++ b/_episodes/19-wrap.md
@@ -30,8 +30,8 @@ turn out to be anything but when we have to explain them precisely.
 
 *   [Pandas](https://pandas.pydata.org) is the home of the Pandas data library.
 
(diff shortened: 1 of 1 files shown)


katrinleinweber:
PS: If you agree, I would send this same PR to [more lessons](https://github.com/search?l=Markdown&o=desc&q=org%3Acarpentries+org%3Aswcarpentry+org%3Adatacarpentry+org%3Alibrarycarpentry+stackoverflow+tagged+NOT+Votes&s=indexed&type=Code)

vahtras:
Good idea

alee:
thanks @katrinleinweber !

katrinleinweber:
You're welcome, and thanks for merging :-)

Issue #444, opened 2019-10-01
Comparing Data Exercise w split vs rpartition


sstevens2:
Any ideas why the Compare Data Exercise in `13-looping-data-sets` uses rpartition vs split?  

From my perspective split is a bit more intuitively named since it splits at every occurrence of the sep where rpartition only splits a the last occurrence of the sep and I had to look it up to figure out what it does.

The code would be identical in the exercise except it would use split.

alee:
Hi Sarah, thanks for the feedback! Here's the commit where it was introduced: https://github.com/swcarpentry/python-novice-gapminder/commit/6b5745a299a060046192e6f8799754bcd649180e

I'm in favor of this change - split is a common function in most languages and Python is one of the few languages that has `*partition` style string methods in its standard library..

There is a minor efficiency gain here for `rpartition` as it scans for the separator starting at the end of the string instead at the beginning but 1) we know we are dealing with small strings and 2) clarity is way more important here..

Would you like to submit a PR for this?

PR #446, opened 2019-10-01
Changing rpartition to split

sstevens2:
re issue #444


sstevens2:
Thanks for the update!  It is great! I really like the description in comments about what the different indexing on split returns.

PR #447, opened 2019-10-01
Online to inline help

sstevens2:
Feel free to ignore this PR but when I think "online" I think looking it up on the internet.  Since this goes along with the function you just created, perhaps "inline" would be better?


vahtras:
oh  github took my dunder-doc as markdown,  I mean 'function.__doc__'

sstevens2:
I'll see if I can think about the rewording a little but I won't have time until next week.  @vahtras Do you think I should close this PR and make an issue instead? or edit on this PR?

vahtras:
keep it - edit if you like

sstevens2:
Finally made these (small) edits.  I liked "builtin" best.  I also edited that sentence to be more accurate about the variable.  Hopefully it is still clear.

PR #455, opened 2019-12-12
Mac_installation_update.md

huajinw:
In installation instructions for Mac OS X, point 3, suggest to install in home directory rather than root directory. I have seen so many problems created by installing in root directory. I am not familiar with Windows, but can assume similar recommendations should be made.

### DIFF
--- a/setup.md
+++ b/setup.md
@@ -20,8 +20,7 @@ you are ready to go as soon as the workshop begins.
 
 ### Windows - [Video tutorial][video-windows]
 
-1. Open [https://www.anaconda.com/download][anaconda-windows]
-   with your web browser.
+1. Open [https://www.anaconda.com/distribution/][anaconda-windows] with your web browser.
 
 2. Download the Python 3 installer for Windows.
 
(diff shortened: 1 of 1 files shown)


vahtras:
It absolutely makes sense. I am not a mac person so I a leave the decision to others, but installing anaconda in a root directory is generally a bad idea.

alee:
LGTM, thank you for submitting this improvement! I have some minor edits for readability and will merge afterwards.

alee:
Sorry for the delay! I just tried installed the latest Anaconda and it looks like it defaults to a local installation on mac os x. I'm still happy to merge this in because it's a good general recommendation but was curious - did you get the root install from the command line installer or the graphical installer?

Thanks again! I'll submit my edits soon against your fork.

Screenshot 2019-12-12 21 39 12

huajinw:
Thanks for editing and merging my pull request. I tried the installation again, and below are the screenshots: initially is wanted to install in root, when I click on “change install location”, it does default “install for me only”, but not that there is an warning at the bottom (2nd screenshot) to indicate I cannot install at the home directory. But when I further click on “install on a specific disk”, I was able to select home directory again. It’s pretty strange, perhaps the installer is not detecting the environment correctly. Nonetheless, I’ve seen people installing at root directory quite often, and thought I would bring it up. Glad you agree that it is a general good practice. The new edits for Windows and Linux installation also look great!

huajinw:
Sorry I realize the screenshots did not show up. I am attaching them again here: 

Screen Shot 2019-12-14 at 9 05 51 AM
Screen Shot 2019-12-14 at 9 06 13 AM

PR #465, opened 2020-02-17
added Slicing riddle

grolimur:
added an exercise on slicing based on Harry Potter and the Chamber of Secrets

### DIFF
--- a/_episodes/02-variables.md
+++ b/_episodes/02-variables.md
@@ -332,6 +332,32 @@ print(ewr_422_yY, 'is', flabadab, 'years old')
 > {: .solution}
 {: .challenge}
 
+> ## Slicing riddle
+>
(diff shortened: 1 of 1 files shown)


alee:
thanks for taking the time to submit this PR! This is a cute example (and I enjoy Harry Potter) but I don't find this to be a great example for slicing; I find it a bit hard to follow with all the string indexing going on. 

I'm open to other @swcarpentry/python-novice-gapminder-maintainers or community thoughts though...

grolimur:
I guess you are right: the example is a little bit complicated.
Thanks for your (nice) feedback.

alee:
Closing in advance of a May deadline to convert to the new Workbench lesson - please reopen there if you'd like us to consider another slicing example. Cheers!

PR #470, opened 2020-04-16
Improve dataframe naming conventions

ndporter:
Proposed updated episode using dataframe naming convention throughout (data_[region]_[indexColumn]) and added bullets pointing out the advantages. The changes would need to be consistent throughout all episodes before being merged to main lesson to demonstrate preferred naming conventions.  Previously dataframes were overwritten and less clearly distinguishable (e.g. `data`).

### DIFF
--- a/_episodes/07-reading-tabular.md
+++ b/_episodes/07-reading-tabular.md
@@ -30,8 +30,8 @@ keypoints:
 ~~~
 import pandas as pd
 
-data = pd.read_csv('data/gapminder_gdp_oceania.csv')
-print(data)
+data_oceania = pd.read_csv('data/gapminder_gdp_oceania.csv')
+print(data_oceania)
 ~~~
 {: .language-python}
 ~~~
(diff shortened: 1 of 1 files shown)


vahtras:
Thanks I agree completely that this is improvement and more in line with good practices. I would welcome a consistent renaming through all episodes. How about about taking it a step further, like data -> data_oceania -> gdp_oceania ?

daviddelene:
I agree, having the same dataframe naming convention would be good.  It would help in understanding what is being plotted and when. Can me more to type; however, the convention is to have meaningful variable names, which this would improve on.

alee:
Sorry we let this languish but the idea is a good one! I've added your changes as well as some minor improvements to my fork at https://github.com/alee/python-novice-gapminder/tree/improve-variable-names and also created issue #470 to track this.

For future reference you should also create a new branch before making changes and submitting a pull request - this will make it easier to keep your fork and this upstream repository properly synchronized. Thanks!

Issue #313, opened 2018-05-08
improve consistency between module and library


alee:
(*pedantry*)

The libraries episode (currently 06) defines a library as a collection of files (modules) but other lesson text refer to specific Python modules as libraries, e.g., the `glob library` should be called the `glob module` instead.

dvalters:
Good point! - I'm going to take a look at this and open a PR soon (if no one is currently working on it)

PR #472, opened 2020-05-01
Consistent use of module and library

msto:
Hi,

I'm making a contribution for instructor training.

Issue #313 requested that the distinction between a "module" and a "library" be more consistent across episodes. After checking all instances of each term in the episode files, I believe there were only a couple instances that could use an update, which I've included. (I also added a stray missing "the").

Please let me know if you need any further changes to address the issue. 

Thanks!


alee:
closes #313

PR #477, opened 2020-05-27
Rephrased dynamical model problem to be more accessible

tkphd:
The logistic map can be quite confusing when first introduced, and the goal of this problem is to practice writing functions, not to stress out the learners with new math concepts. These changes introduce the logistic map equation (which had been hidden in the "solution" area), describe what it does, and rename several variables to be less abstract. This better-grounded version of the problem statement ought to be more accessible. A print statement has also been added, with representative output, that demonstrates the utility of the logistic map, just for good measure.

These changes are not expected to break anything, but the code in PR #344 may be in conflict.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ before_install:
   - sudo apt-key adv --keyserver keyserver.ubuntu.com --recv-keys E084DAB9
   - echo "deb https://cran.rstudio.com/bin/linux/ubuntu trusty/" | sudo tee -a /etc/apt/sources.list
   - sudo apt-get update -y
-  - sudo apt-get install -y r-base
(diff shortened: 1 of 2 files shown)


vahtras:
Thank for the input. I think your PR improves the presentation and I see no issues.

tkphd:
Thanks @vahtras, happy to contribute.
The second commit addresses a build failure on Travis-CI due to a configuration problem blocking dependencies from installing. After the patch, [it does indeed pass](https://travis-ci.com/github/tkphd/python-novice-gapminder/builds/169467086).

I know the second commit is out of scope from the PR's headline, but it's a trivial change that fixes a breakage, so I hope it will be merged as well.

PR #483, opened 2020-06-22
Programming Style: PEP8 -> pycodestyle

tamachado:
I made two proposed changes to the section about programming style:

1) The `pep8` application for checking your code is no longer maintained. The most recent version is now called `pycodestyle` to avoid confusion with the name of the standard itself. This PR fixes this.

2) I thought the mention of the alternative Google style guidelines was sort of confusing (which I do not believe to be a strict superset of PEP8). So I tried to make the language a bit clearer that there are just multiple standards that one could choose to follow.


vahtras:
Thanks! I agree that names of applications that are out of date should be fixed, but I would not want to remove the reference to Google/yapf. Possibly one could mention a few more, like black, and in particular if any are useful in the context of a notebook (which I am not aware of...). Thoughts from any others?

tamachado:
We could also mention autopep8, and I guess there's a [jupyter wrapper for black/yapf](https://github.com/dnanhkhoa/nb_black) which I haven't tried.

daviddelene:
Yes, I would leave in reference to Google/yapf since this is a different style.  It good to point out that there are several styles, not just one.  It would that people follow an existing style instead of just making up there own.

PR #485, opened 2020-06-30
Fixed broken links to other episodes

schmoelder:
I found two broken links in the documentation. I tried to fix them by using `page.root` instead of the relative link. I checked it locally (after running make serve) and it seemed to work. Hopefully it also works for the Github pages.


souravsingh:
LGTM

PR #489, opened 2020-08-19
Add a table for arithmetic operators

akrinos:
I was recently teaching this lesson and a student was confused about the arithmetic operators (younger student who likely had only, for example, used a calculator with the standard division sign). For that reason, I added a small table with the arithmetic operators before the first operation is performed, just for the students to have for reference. I figured that a table would add minimally to the lesson time while also providing the background for students who might be unfamiliar.

### DIFF
--- a/_episodes/02-variables.md
+++ b/_episodes/02-variables.md
@@ -109,6 +109,16 @@ NameError: name 'last_name' is not defined
 
 *   We can use variables in calculations just as if they were values.
     *   Remember, we assigned the value `42` to `age` a few lines ago.
+*   Python has a set of built-in arithmetic operators for calculations. 
+
+| Operation     | Operator | 
+|-------------- | -------- |
+| Addition      |    +     |
+| Subtraction   |    -     | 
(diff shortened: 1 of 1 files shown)


vahtras:
Thank you for your suggestion. Although I appreciate your effort to clarify notation for novices, I don't think this is the place for such a table. The focus on this section is variable assignment and my feeling is that this table breaks the narrative in this context, and is not required to explain this example with only addition. It would make more sense to have something similar in episode 3 where we have more use of these operators including discussions on how they mean different things for different data types. I leave the question open for other maintainers.

alee:
Thanks for the PR! Apologies for the long delay in response & review, but I agree overall with @vahtras assessment that it seems a bit out of place here in this episode where the primary goal is to introduce the idea of variables and how to manage and store data in programs. Perhaps in episode 3?

In any case, I'm closing this PR in preparation for a Carpentries-wide lesson infrastructure upgrade. Please feel free to resubmit or open an issue to gather community input on this suggestion.

PR #495, opened 2020-08-25
A note on jupyter syntax highlighting

emmanueliarussi:
As I was teaching with episodes across this lesson, I noticed that syntax highlighting is never mentioned. It may be helpful to bring it up explicitly at this point, since it helps to pay attention and correct mistakes related to variable types.

### DIFF
--- a/_episodes/02-variables.md
+++ b/_episodes/02-variables.md
@@ -26,6 +26,7 @@ keypoints:
 *   The variable is created when a value is assigned to it.
 *   Here, Python assigns an age to a variable `age`
     and a name in quotes to a variable `first_name`.
+*   Jupyter performs automatic syntax highlighting, showing strings in red. 
 
 ~~~
 age = 42


alee:
Thanks for the PR! And apologies for the late attention given to it. I'm open to this change, but I am a bit hesitant to list the exact color (red) used in the current Jupyter notebook's syntax highlighting - this seems like something that could change in future versions. Perhaps we can explain what syntax highlighting is, instead? Something like

```
Jupyter notebooks perform automatic syntax highlighting on your code, so things like strings, function names, and reserved Python keywords will be highlighted with different colors to help you notice them.
```

alee (commented):
Thanks for the PR! And apologies for the late attention given to it. I'm open to this change, but I am a bit hesitant to list the exact color (red) used in the current Jupyter notebook's syntax highlighting - this seems like something that could change in future versions. Perhaps we can explain what syntax highlighting is, instead? Something like

```
Jupyter notebooks perform automatic syntax highlighting on your code, so things like strings, function names, and reserved Python keywords will be highlighted with different colors to help you notice them.
```

alee:
Closing in preparation for a Carpentries-wide lesson infrastructure upgrade - please reopen again later if so desired (taking feedback into account). Thank you!

PR #496, opened 2020-09-22
Add space around <div> tags

zkamvar:
## Problem

Github's markdown parser gets confused because [some of the code fences inside div elements are not parsed as markdown](https://github.com/swcarpentry/python-novice-gapminder/blob/a2fbcf1c9e8aa0b770f93cd5012d9d4cfbd821ca/_episodes/01-run-quit.md#markdown-does-most-of-what-html-does)

screen capture showing incorrectly parsed markdown where div tags are included in code blocks

## Solution

This PR adds blank lines around div tags to avoid mistakes in parsing for markdown engines that are based on the [strict markdown spec for including HTML elements](https://daringfireball.net/projects/markdown/syntax#html) states:

> The only restrictions are that block-level HTML elements — e.g. `<div>, <table>, <pre>, <p>`, etc. — must be separated from surrounding content by blank lines

### DIFF
--- a/_episodes/01-run-quit.md
+++ b/_episodes/01-run-quit.md
@@ -284,36 +284,45 @@ more details, then see the [official notebook documentation][jupyterlab-notebook
 
 <div class="row">
   <div class="col-md-6" markdown="1">
+    
 ~~~
 *   Use asterisks
 *   to create
 *   bullet lists.
 ~~~
+
   </div>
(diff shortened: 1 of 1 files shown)


alee (approved):
I checked this out locally and it fixed the issue, thank you for the PR!

zkamvar:
Thank you for being so quick to respond!

Issue #370, opened 2019-02-13
Variables copied by reference not mentioned in Lesson 3


dalonsoa:
In Lesson 3, it is stated that "Variables only change value when something is assigned to them" in one of the headings. That is not correct when the variables are copied by reference to another variable: When the original variable is changed, the second one changes, too. 

This is easily seen with lists:

```python
a = [1, 2, 3]
b = a
a[2] = 4
print(a)
print(b)
```

The output in both cases is [1, 2, 4] even though we didn't changed variable `b` explicitly. This effect can be very confusing to new users, getting unexpected results when they didn't change their variables, and I think it deserves an explanation somewhere within this lesson. 

I'm happy to contribute to that part if it seems like a reasonable addition. 

Best wishes, 
Diego

dalonsoa:
I've seen that copy by reference is mentioned, just as an example, in Lesson 11: Lists, but I think it should also be discussed here on its own or, at least, clarify this statement is not entirely true and that will be revisited in Lesson 11.

ntmoore:
I agree that copy vs point to the original is a confusing and subtle point.  I wonder if this would be a useful exercise in episode 3 and/or 11.

dvalters:
Agreed - just finished teaching this course and this was one of the issues that came up. The fact that lists have this behaviour (and Pandas dataframes) leads to a confusing realisation when students start using these data structures, having previously being told the contrary case early on. There is a related exercise in Episode 11: "Copying (or Not)", but could be introduced early on.

davidherbert2:
As a new instructor, I would like to take a look at this issue as a checkout task during October 2020.

alee:
#499 included changes and text improvements for this issue but stalled on feedback:

https://github.com/swcarpentry/python-novice-gapminder/pull/499#pullrequestreview-512370133

PR #499, opened 2020-10-05
Wording to address #370

davidherbert2:
Have added hopefully clarifying examples to the section on variable assignment, drawing attention to the issue raised in #370 that the statement in the section title is not true in all cases.  Refers to the example in episode 11 to put the issue in a list context.

### DIFF
--- a/_episodes/03-types-conversion.md
+++ b/_episodes/03-types-conversion.md
@@ -220,7 +220,7 @@ first is 2 and second is 5
     creates a new value, and assigns it to `second`.
 *   After that, `second` does not remember where it came from.
 
-Sometimes, however (and this can happen with Python list types, described in more detail in Episode 11:Lists), more than one variable can point to the same storage location.  Think of a cell in a spreadsheet which depends on another cell.  Updating the latter will automatically update the former.  The dependent cell is actually a **reference** to the original, not a distinct copy of it.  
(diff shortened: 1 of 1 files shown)


alee (commented):
Thanks for the PR! I think this is an important improvement to the lesson and clarifies things nicely. However, (and apologies for the pedantry), `b` is not really a copy of `a`. It's being assigned to the value of a, the canonical `int` object `1`. You can't mutate / modify an int object, or a string object, or a tuple object - they are guaranteed to be immutable.

However, if you assigned `a` to a mutable object like a list, map, set, or a custom class that holds some variable references to other value objects (e.g., numbers, strings, tuples), you could dig into that mutable object and change one of those internal references to point at a different value object.

You can test this by using the `id` function, e.g.,

```
a = 1
b = 2
id(a)
id(b)
b = a
id(b)
```

Python is a `call-by-object-reference` language, where everything is an object, and what really matters is if you are assigning a mutable object to a variable name, or an immutable object to a variable name. If you assign a mutable object, then you have the ability to change some of its mutable references, but something like:

```
a = [3]
b = a
b = [4]
```
does not change `a` for example. 

Would you be willing to revise this PR to include some of this nuance? I'd be happy to help edit the PR myself but don't have much bandwidth until the end of this month or the beginning of November likely...

Thanks again for taking the time to help us improve this lesson!

BillYi1990:
Agreed with alee's comments that b is not really a copy of a, it should be assigned to the current value of a. #370 #396 #499
`a = [1, 2, 3]`
`b = a`
`print(a)`
`print(b)`
`a[2] = 4`
`print(a)`
`print(b) # assigning current value of a to b `

alee:
Closing this PR in preparation for a Carpentries-wide lesson infrastructure upgrade. Please feel free to resubmit or we can continue to gather community input on this suggestion in #370

PR #509, opened 2020-10-21
Correct exercise output

unode:
The code cell and output are out of sync.
The last line also includes quotes which doesn't happen when using print()

Please delete this line and the text below before submitting your contribution.

Credit to @tobyhodges for bumping into it during a course.


alee:
thanks for the PR @unode and catching this @tobyhodges !

PR #512, opened 2020-12-05
Change the variable name "sum" to "summatory"

AbrahamAvelar:
As "sum" is a built-in function, it is better to name this variable something different.

### DIFF
--- a/_episodes/12-for-loops.md
+++ b/_episodes/12-for-loops.md
@@ -320,23 +320,23 @@ print(total)
 > The result should be `[1, 3, 5, 10]`.
 >
 > ~~~
-> cumulative.append(summatory)
+> cumulative.append(total)
 > for number in data:
 > cumulative = []
-> summatory += number
-> summatory = 0
+> total += number
+> total = 0
 > print(cumulative)
 > data = [1,2,2,5]
 > ~~~
 > {: .language-python}
 > > ## Solution
 > > ~~~
-> > summatory = 0
+> > total = 0
 > > data = [1,2,2,5]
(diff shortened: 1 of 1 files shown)


alee (changes requested):
Thanks for taking the time to improve our lesson with this PR! I agree completely that we should not use any of the [built-in methods](https://docs.python.org/3/library/functions.html) as variable names. That said, `summatory` is a bit arcane, can we use something like `cumulative_sum` or `total` instead?

vahtras:
A practice that is not uncommon is to append an underscore to a variable name that conflicts with a built-in, e.g. in this case `sum_`.

AbrahamAvelar:
I agree that cumulative_sum or total may be better. sum_ is good too but I think may be confussing to novices. Anyways I believe the important think is to change the original name from the lesson as it is confussing, even more as in the method sum() is used in the "Writing Functions" episode. When running sequencially, the error comes from python trying to use it as a variable name.

alee:
> I agree that cumulative_sum or total may be better. sum_ is good too but I think may be confussing to novices. Anyways I believe the important think is to change the original name from the lesson as it is confussing, even more as in the method sum() is used in the "Writing Functions" episode. When running sequencially, the error comes from python trying to use it as a variable name.
> How do we get the change in the SC lesson?

If you're OK with `cumulative_sum` or `total` I'd be happy to merge after renaming `summatory` -> `cumulative_sum` - you can do this by adding another commit to your PR branch. Let us know if you need help with it, and thanks again for the PR!

alee (approved):

Issue #115, opened 2016-10-26
Should the final point on functions in episode 4 be moved to fit with the other information about functions?


lo5an:
In episode 4, "Every function returns something."  might more naturally fit with the other points about functions, rather than following the section on help and debugging.

PR #513, opened 2020-12-05
reorder 2 sections of the built-in function lesson

eldobbins:
This PR addresses issue #115 . 
> In episode 4, "Every function returns something." might more naturally fit with the other points about functions, rather than following the section on help and debugging.

This PR:

* Moves "every function returns something" to directly after an example that uses print() to keep similar examples together
* Moves the Notebook help functions before the error messages
* Clarifies the text regarding Notebook help functions


alee (approved):
Thanks for taking the time to improve the lesson! LGTM

Issue #461, opened 2020-01-15
"Writing Functions" - inconsistent question and solution formatting


stb2145:
I'd like to bring attention to the inconsistent formatting in the "Writing Functions" episode in the "Plotting and Programming in Python" lesson (http://swcarpentry.github.io/python-novice-gapminder/14-writing-functions/index.html). From a novice's perspective having a more consistent formatting when introducing questions and providing solutions would be very helpful - a simple fix would be to number the questions/solutions appropriately. 

Before I noticed that the question/answer formatting could be improved, I was struggling to understand what was trying to be explained in the "Order of Operations" block. In this case, numbering the questions and associated answers appropriately would've gone a long way in trying to understand what exactly was being asked and the solutions provided (perhaps the wording int he Q/A in this block could be more clear as well...).

The "Encapsulating Data Analysis" block has a very clear question-to-answer format where they are simply numbered in a consistent manner. As mentioned, a simple fix would be to follow that formatting throughout the episode!

alee:
fixed in #514 - thanks @eldobbins !

PR #514, opened 2020-12-07
edits to standardize the format of exercises

eldobbins:
This PR is to address issues described in #461, which describes an exercise whose formatting could be improved.  While working on that exercise, I found a couple more with formatting oddities, so fixed them too.

* `Order of Operations`
  * added numbering to match questions with solutions
  * reordered the questions so that the function parse_dates() is introduced before it is asked about
  * added a 3rd question about the `None` returned by parse_dates()
* removed the Python block formatting from `Calling by Name` because I think that section is part of the questions not part of the code
* added spaces to the items list in `Encapsulating Data Analysis` so the Markdown is parsed correctly
* `Encapsulate of If/Print Block`
  * fixed typo in the title
  * added and removed spaces from code examples to be more PEP8

I'm not satisfied with the wording of that last exercise.  It asks 2 questions and then solves them before asking 2 more whose answers are in the Solutions block.  I see that the exercise author is trying to chain concepts, but I'm not 100% sure it's successful. Please let me know if you want me to reformulate that exercise as part of this PR

### DIFF
--- a/_episodes/16-writing-functions.md
+++ b/_episodes/16-writing-functions.md
@@ -201,24 +201,7 @@ result of call is: None
 
 > ## Order of Operations
 >
-> The example above:
->
-> ~~~
-> result = print_date(1871, 3, 19)
-> print('result of call is:', result)
-> ~~~
-> {: .language-python}
->
-> printed:
-> ~~~
-> 1871/3/19
-> result of call is: None
-> ~~~
-> {: .output}
->
-> Explain why the two lines of output appeared in the order they did.
->
-> What's wrong in this example?
+> 1. What's wrong in this example?
 > ~~~
 > result = print_date(1871,3,19)
 >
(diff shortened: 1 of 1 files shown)


alee:
Reformulation of the exercise would be welcome, but we can have a separate PR for that - thanks again!

eldobbins:
Maybe we could close the original issue #461 ?

eldobbins:
I just checked the formatted rendering of the changes I made, and it's all screwed up.  I'm going to try and fix that.  I still have the branch.   Can this PR be reopened or should I start a new one?

eldobbins:
OK.  I've fixed the formatting problems.  The fixes are in the original branch.  Please let me know how I should submit it.

eldobbins:
Issued a new PR to fix this rendering problem.

PR #517, opened 2020-12-19
reformatting 2 exercises so they render correctly

eldobbins:
When I submitted PR #514, I neglected to check the rendering of the exercises I had edited.  Turns out, I was missing some blank lines so the code blocks were all messed up.

While I was fixing that, I figured out how to intersperse code blocks with markdown lists so that the list items increment correctly (instead of resetting to 1 after the code block). This was a problem with another exercise, so I fixed that too.

The 2 exercises modified are "Order of Operations" and "Encapsulating Data Analysis"


alee (approved):

alee:
Doh! Thanks for catching that, I should have looked more carefully at the rendered version locally in my last review.

Issue #473, opened 2020-05-01
Briefly mention methods when using DataFrame.describe


charlesincharge:
Just taught this lesson to programming beginners, and they got tripped up on the different syntax of methods. It would help to have something like:
> Some data types have _methods_ associated with them, which are like functions that operate on the associated value. Methods are called similarly to a function with: <variable_name>.<method_name>(arguments)

When the first method (`DataFrame.describe`) gets used in https://swcarpentry.github.io/python-novice-gapminder/07-reading-tabular/index.html, there isn't a explicit mention of methods.

aayux:
+1

Perhaps it would help to make a clear distinction made between data members and member functions before jumping into the examples.

Even without a complete understanding of classes or objects &mdash; and without getting into design philosophy &mdash;  it would be useful to know how the two constructs (data and methods/functions) differ. IMO a brief one or two line explanation should suffice.

RMeli:
@charlesincharge @aayux Are you working on this issue or can I fix it as part of the checkout process?

aayux:
It's fine by me. This has been open for a while and there are no PRs so I think it should be okay.

charlesincharge:
@RMeli I'm not working on this actively. You/anyone else should feel free to fix this. Thanks!

eldobbins:
I recently taught this, and included a brief intro to methods in Episode 4.  I used strings as an example because they are simple and have many methods associated with them.

I will submit a PR of what I did.

PR #341, opened 2018-09-23
add challenge that introduces the dir command

ntmoore:
I think the dir function/command is very useful for finding out what functions are available for a given data object.  Given that this is an optional side-project, I included an introduction to the topic as a challenge.

Note, I totally waved my hands and glossed over the "double underscore" data/functions that dir returns.  Someone with more expertise in python should (1) fix my terminology and (2) decide what level of description is necessary for those components.

further modifications (PR's) on this are requested!

PR #518, opened 2020-12-22
Introducing methods in episode 4, and using them in episodes 7 and 8

eldobbins:
This PR is to address issue #473 which requests an introduction to methods be added before methods are used in the pandas episodes.  This PR:

* adds an example of methods using strings in Episode 4
* updates some text in Episode 7 - the first time I could find where a method is used
* changes the dir() exercise in Episode 8 (PR #341) to use strings instead of a list. This provides continuity and avoids using lists, which haven't been introduced yet.

I would like to add a couple changes to this PR, but I wanted to check first, because they aren't directly related. These are all in Episode 7

* ~~A reference to `math.pi` is made when introducing `DataFrame.columns`. I would remove that because I don't see any other references to `math`, which makes it confusing to me.~~  I just noticed the libraries lesson that covers that.  Should have seen that before.
* Transposing is introduced as `data.T` and it is described as being like `DataFrame.columns`.  I would substitute `data.transpose()` because `T` isn't like `columns`.  `columns` is an attribute and according to the [pandas docs](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.transpose.html?highlight=transpose#pandas.DataFrame.transpose), "The property T is an accessor to the method transpose()."  I'm not sure I could explain what that means, so I'd like to use the function instead. `T` could then be described as a "shortcut".
* Methods are described as "members" and attributes like `columns` are described as "member variables". I've never seen that terminology before. Does it come from another language. like C++?  I would like to use the terms "attribute" and "methods" exclusively since those are used in the [DataFrame pandas docs](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.html).

Please, let me know what you think about those 3 additional changes.

### DIFF
--- a/_episodes/04-built-in.md
+++ b/_episodes/04-built-in.md
@@ -137,6 +137,44 @@ round(3.712, 1)
 ~~~
 {: .output}
 
+## Functions attached to objects are called methods
+
+* Functions take another form that will be common in the pandas episodes.
+* Methods have parentheses like functions, but come after the variable.
+* Some methods aren't used often, and are marked with double underlines.
+
+~~~
+my_string = 'Hello world!'  # creation of a string object 
+
+print(len(my_string))       # function with the string as an argument
+
+print(my_string.__len__())  # method acting upon the string object
+~~~
+{: .language-python}
+
+~~~
+12
+12
+~~~
+{: .output}
+
+* You might even see them chained together.  They operate left to right.
+
+~~~
+print(my_string.isupper())          # Not all the letters are uppercase
+print(my_string.upper())            # This capitalizes all the letters
+
+print(my_string.upper().isupper())  # Now all the letters are uppercase
+~~~
+{: .language-python}
+
+~~~
+False
+HELLO WORLD
+True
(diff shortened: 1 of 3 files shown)


alee:
This is great! I agree with all of the additional changes: transpose vs pandas.T which uses the Python `@property` decorator to provide "please-drive-through" access to the transposed rendition of the given data frame and harmonizing language to match the existing pandas docs. Members / member variables are commonly used to describe these constructs in object oriented programming (e.g., https://docs.python.org/3/tutorial/classes.html) but I like the terms `methods` and `attributes` - they clearly distinguish between the function and data "members" of a class.

I'm happy to merge this in as is if you'd like to make those changes in a separate PR, but if you'd like to add more to this PR that would be fine too - just let me know!

Sorry for the late review on this, the holidays and a busy start to our year didn't leave me much time to look into these. Thanks for all your help improving the lesson!

alee (approved):

vahtras (approved):

PR #519, opened 2021-01-25
Improve solution to exercise in For Loop episode

santisoler:
Improve the explanation of the solution for the "Identifying Variable
Name Errors" exercise in the For Loops episode (12). Rearrange the
expected errors to mimic the order in which they they would appear as we
solve them. Add missing expected error for undefined variable `a`.

### DIFF
--- a/_episodes/12-for-loops.md
+++ b/_episodes/12-for-loops.md
@@ -365,8 +365,11 @@ print(total)
 > ~~~
 > {: .language-python}
 > > ## Solution
-> > The variable `message` needs to be initialized and Python variable names are case sensitive: `number` and `Number`
-> > refer to different variables.
+> > Python variable names are case sensitive: `number` and `Number` refer to
+> > different variables.
+> > The variable `message` needs to be initialized as an empty list.
+> > We want to add the string `"a"` to `message` and not the undefined variable
+> > `a`.
 > > ~~~
 > > message = ""
 > > for number in range(10):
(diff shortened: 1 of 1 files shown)


vahtras (approved):
Looks good tome

santisoler:
I'm glad to be helpful. I was wondering if there's a way to put each one of the sentences I created as bullet items. I tried to do so, but the text was looking odd (fonts in purple and as they were inside the code block):

image

If there isn't an easy way to do that, I'm ok with how it looks right now.

alee:
Thanks for the PR @santisoler ! In order to get it to render properly add a newline after your bulleted items, e.g.,

```
> {: .language-python}
> > ## Solution
> > - Python variable names are case sensitive: `number` and `Number` refer to different variables.
> > - The variable `message` needs to be initialized as an empty list.
> > - We want to add the string `"a"` to `message` and not the undefined variable `a`.
> > 
> > ~~~
...
```

> I'm glad to be helpful. I was wondering if there's a way to put each one of the sentences I created as bullet items. I tried to do so, but the text was looking odd (fonts in purple and as they were inside the code block):
> 
> ```
> > {: .language-python}
> > > ## Solution
> > > * Python variable names are case sensitive: `number` and `Number` refer to
> > > different variables.
> > > * The variable `message` needs to be initialized as an empty list.
> > > * We want to add the string `"a"` to `message` and not the undefined variable
> > > `a`.
> ```
> 
> image
> 
> If there isn't an easy way to do that, I'm ok with how it looks right now.

santisoler:
Thanks @alee for the tip! I just changed the sentences to a bullet list, which I think looks better.

I've also noticed that I messed with the type of the `message` variable: it's supposed to be a string and not a list! Already fixed that!

So feel free to merge if you agree.

PR #527, opened 2021-03-01
06-libraries.md: randint vs. randrange

maxim-belkin:
`randint` and `randrange` are different methods/functions, so I propose to remove the part that says that one is an alias for another.


alee:
thanks @maxim-belkin (and for all the other PRs! :tada: )

From https://docs.python.org/3/library/random.html#random.randint `randint(a, b)` is called out as equivalent to `randrange(a, b+1)` for a given `a, b` but I agree that we don't need to mention that they are aliases (or just say exactly what the Python 3 docs say)

alee (approved):

PR #530, opened 2021-03-03
11-lists: add missing parenthesis

maxim-belkin:
Well, a parenthesis is missing...


alee (approved):
doh, thanks Maxim! My background in Scheme and LISP-y languages is ashamed of myself

PR #531, opened 2021-03-04
Fix formatting typo

PabRod:
The modulo operator `%` was rendered differently than `/` and `//`.


vahtras:
Thanks, good catch!

Olav

Issue #516, opened 2020-12-16
Add directions for alt launching using anaconda-navigator


sstevens2:
The way the launching section is written now it sounds a bit like you can't launch a jupyter lab without using the terminal on a mac or jupyter prompt for windows. Might be good to offer alt directions for how you can launch it with anaconda navigator instead.  Learners often seem to prefer that option during the workshops I've taught.

K-Meech:
I can add instructions for launching via anaconda navigator - I'll make a PR tomorrow!

alee:
added in #533

PR #533, opened 2021-03-13
add anaconda navigator instructions in 'Running and Quitting' episode

K-Meech:
Adds instructions for opening JupyterLab via anaconda navigator - for issue: https://github.com/swcarpentry/python-novice-gapminder/issues/516

I tested this on a new Anaconda installation on Windows. I believe the Mac instructions are also correct, but it would be good to check with a mac user that:
- the instructions also work for them
- the anaconda navigator interface looks similar to the screenshot in the lesson


paultgriffiths:
Finding things via Spotlight [can be temperamental ](https://apple.stackexchange.com/questions/62715/applications-dont-show-up-in-spotlight) so a belt-and-braces approach (if required) might be to include the option to Command-Shift-G in Finder and then type /Applications to be sure to go to the Mac Applications folder.

alee (approved):
Thanks @K-Meech ! I made some minor edits, if you think they look OK I'm happy to merge in (so long as no other @swcarpentry/python-novice-gapminder-maintainers have any objections)

alee:
Thanks for the heads up @paultgriffiths - do you think the current set of instructions is sufficient? I don't want to overwhelm with too many options at this point and would prefer to keep things as succinct as possible.

> Finding things via Spotlight [can be temperamental ](https://apple.stackexchange.com/questions/62715/applications-dont-show-up-in-spotlight) so a belt-and-braces approach (if required) might be to include the option to Command-Shift-G in Finder and then type /Applications to be sure to go to the Mac Applications folder.

K-Meech:
> Thanks @K-Meech ! I made some minor edits, if you think they look OK I'm happy to merge in (so long as no other @swcarpentry/python-novice-gapminder-maintainers have any objections)

All looks good to me @alee!

paultgriffiths:
> Thanks for the heads up @paultgriffiths - do you think the current set of instructions is sufficient? I don't want to overwhelm with too many options at this point and would prefer to keep things as succinct as possible.
> 
> > Finding things via Spotlight [can be temperamental ](https://apple.stackexchange.com/questions/62715/applications-dont-show-up-in-spotlight) so a belt-and-braces approach (if required) might be to include the option to Command-Shift-G in Finder and then type /Applications to be sure to go to the Mac Applications folder.

Definitely sufficient - on most modern Macs, the Spotlight database updates pretty quickly, so Spotlight should find things after they've been installed.  If not, going to Applications will definitely work.

PR #535, opened 2021-03-15
pep8 tool was renamed as pycodestyle

villares:
ops! this changes the URL to the tool mentioned, tha was renamed
pep8 -> pycodestyle


alee (approved):
thanks!!

Issue #532, opened 2021-03-11
Ways of accessing DataFrame elements, rows, columns, and subsets


maxim-belkin:
This isn't really an issue, but rather me trying to summarize ways one can access various subsets of a DataFrame in Pandas. Perhaps, it would be of benefit to someone or inspire someone to improve the lesson... :)))

1. Access a single column
    ```
    # by name
    data["col_name"]   # as a Series
    data[["col_name"]] # as a DataFrame

    # by name using .loc
    data.T.loc["col_name"]  # as a Series
    data.T.loc[["col_name"]].T  # as a DataFrame

    # Dot notation (Series)
    data.col_name

    # by index (iloc)
    data.iloc[:, col_index]   # as a Series
    data.iloc[:, [col_index]] # as a DataFrame

    # using a mask
    data.T[data.T.index == "col_name"].T
    ```
2. Access a single row
    ```
    # by name using .loc
    data.loc["row_name"] # as a Series
    data.loc[["row_name"]] # as a DataFrame

    # by name
    data.T["row_name"] # as a Series
    data.T[["row_name"]].T as a DataFrame

    # by index
    data.iloc[row_index]   # as a Series
    data.iloc[[row_index]]   # as a DataFrame

    # using mask
    data[data.index == "row_name"]
    ```
3. Access an individual DataFrame element
    ```
    # by column/row names
    data["column_name"]["row_name"]         # as a Series

    data[["col_name"]].loc["row_name"]  # as a Series
    data[["col_name"]].loc[["row_name"]]  # as a DataFrame

    data.loc["row_name"]["col_name"]  # as a value
    data.loc[["row_name"]]["col_name"]  # as a Series
    data.loc[["row_name"]][["col_name"]]  # as a DataFrame

    data.loc["row_name", "col_name"]  # as a value
    data.loc[["row_name"], "col_name"]  # as a Series. Preserves index. Column name is moved to `.name`.
    data.loc["row_name", ["col_name"]]  # as a Series. Index is moved to `.name.` Sets index to column name.
    data.loc[["row_name"], ["col_name"]]  # as a DataFrame (preserves original index and column name)

    # by column/row names: Dot notation
    data.col_name.row_name
    
    # by column/row indices
    data.iloc[row_index, col_index] # as a value
    data.iloc[[row_index], col_index] # as a Series. Preserves index. Column name is moved to `.name` 
[...]

PR #537, opened 2021-04-07
Episode 8: Add 'Many Ways of Access' challenge.

vinisalazar:
This is an improvement based on the suggestions from #532.

I added a Challenge asking learners to suggest different ways of doing access operations on a DataFrame. I used the notes from #532 as answer to the Challenge.

Thank you for any assistance or feedback you can provide.

Best,
V

### DIFF
--- a/_episodes/08-data-frames.md
+++ b/_episodes/08-data-frames.md
@@ -509,6 +509,155 @@ data.groupby(wealth_score).sum()
 > {: .solution}
 {: .challenge}
 
+> ## Many Ways of Access
+>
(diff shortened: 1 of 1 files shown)


alee (approved):
Thanks for the PR! I'll double check the formatting later tonight and merge if there's no objections from other community members.

alee (approved):
Thanks for submitting, looks great! Apologies for the delayed review.

vinisalazar:
Thank you for the review @alee! Happy to contribute.

PR #543, opened 2021-04-21
Add callout block referring to Python docs

vinisalazar:
Hi,

This PR adds a callout block to the end of the episode that links to the Python documentation.

There are other mentions of the Python docs throughout the lesson, but I considered that the Built-in Functions topic would be a good place to insert a more complete reference to the docs page. I hope this change is helpful to the lesson.

Best,
Vini


alee (approved):
thanks! This looks good to me.

Issue #501, opened 2020-10-07
updating Python setup instructions so that "check Add Anaconda to Path" is recommended


rbavery:
These setup instructions more clearly direct learners to add Anaconda to their PATH using the most recent language in the Anaconda installer: https://carpentries.github.io/workshop-template/

Versus the current instructions: http://swcarpentry.github.io/python-novice-gapminder/setup

Are the maintainers open to a PR that copies over the Python setup instructions from the workshop template, including the instructional video?

alee:
That would be great! Apologies for the much delayed response

vinisalazar:
Hi,

I submitted PR #545 that attempts to address this. I updated `setup.md` according to the workshop template instructions.

Best,
V

PR #545, opened 2021-04-26
setup.md: Update Python installation instructions

vinisalazar:
Hi,

Please see #501 for discussion.

If merged, this PR will:
- Add the `_includes/python_install.html` file with setup instructions, the same present in [carpentries/workshop-template](https://github.com/carpentries/workshop-template/blob/gh-pages/_includes/install_instructions/python.html) repository. This file contains embedded video instructions for each operating system.
- Include said file in the `setup.md` (and delete previous instructions). I also moved the "Getting the Data Section" up so it does not stay hidden below the Python setup section.
- Update the `assets/css/lesson.scss` file according to the [carpentries/workshop-template](https://github.com/carpentries/workshop-template/blob/gh-pages/assets/css/lesson.scss) repository.

I do know that altering static files may be a bit risky, but that was needed so the Python setup instructions block would look right. I rendered the lesson locally on OS X and could not find any problems.

Best,
V

### DIFF
--- a/_includes/python_install.html
+++ b/_includes/python_install.html
@@ -1,0 +1,112 @@
+{% comment %}
+
+Remove the third paragraph if the workshop will teach Python
+using something other than the Jupyter Notebook. Details at
+
+https://jupyter-notebook.readthedocs.io/en/stable/notebook.html#browser-compatibility
+{% endcomment %}
+<div id="python">
+
+  <p>
+    <a href="https://python.org">Python</a> is a popular language for
+    research computing, and great for general-purpose programming as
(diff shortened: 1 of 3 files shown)


rbavery (commented):
Looks solid, thanks @vinisalazar

vinisalazar:
Pinging @alee, could we get a position from the maintainers regarding this one?

Thank you for any assistance you can provide,
Vini

vahtras:
Looks good to me!

alee (approved):
LGTM, tested locally with `make docker-serve`

Thanks, and sorry for the delayed review!

Issue #538, opened 2021-04-09
In 16-writing-functions, print_date example needs revising


ognancy4life:
In lesson 16: Writing Functions, there is a question regarding "Order of Operations" using the print_date function. This is a good question, but because print_date has already been defined earlier in the lesson, the students will not get the desired error. I recommend using a different example function here to teach order of operations. Thank you!

alee:
Excellent point, thank you for raising it!

http://swcarpentry.github.io/python-novice-gapminder/16-writing-functions/index.html#order-of-operations

AlexT97:
One suggestion with Order of Operations, I think it might be clearer for students looking through this to have multiple examples to work with.  The print_date is probably okay to work with but I would recommend something a little bit more basic to go alongside. One thing that might be useful would be to have another full example of a function definition, like @ognancy4life suggested, but using mathematical expressions in such a way that if the arguments to the function are entered incorrectly, the mathematical result will be obviously incorrect.  This way, in the section itself, we could explain what the expected result is and why, if we enter the arguments in the wrong order, the function will return the wrong example.  In addition, it might be helpful to explain why an error occurs with incorrect order of operations in words as well as in the example.  Approaching the issue in multiple vectors of understanding is usually really helpful for students to better grasp what's going on.

perlman:
My PR (#546) was an effort to change this with the minimal changes.

The example @AlexT97 provides would certainly work, and might be superior in that there would be less to type in if a student tries to do so.  I guess the question is which additional concepts would be less confusing: using parentheses to order mathematical operations, or string building?

I don't know. :)

PR #546, opened 2021-04-27
Replace print_date function in episode 16 question

perlman:
As noted in #538, the function `print_date` was previously used in the episode, leading to a student to potentially fail to experience the expected error if they type the code into their current python session.

This updated question will use `print_time` instead. This is a very simple substitution which I should be as accessible as  `print_date`.


alee (changes requested):
thanks for the PR! I've added a few minor comments to be fixed and then I think this will be ready to merge.

Issue #539, opened 2021-04-12
Jupyter is not the only way to run python


jatonline:
In recent (non-Carpentry) Python courses, we have come across learners that have experience with Python and using JupyterLab or Jupyter Notebooks, however are unaware that you can just run a Python script from the command line. We have observed that this has led to some confusion when they've been working with others who use script files.

I'm not for a second suggesting changing the way the lessons operate, as JupyterLab is definitely great for teaching (and in any case, this matter has been discussed before in #15). However, I wonder whether it would be helpful to introduce a little earlier in the lesson that you can run Python in lots of different ways, either directly from a command line, or through Jupyter, and we are going to use Jupyter for the lesson.

This is already mentioned on line 210 of the first episode (https://github.com/swcarpentry/python-novice-gapminder/blame/59d5eeec65a91143e74910d39e7a88e99219de41/_episodes/01-run-quit.md#L210), however given that the question the episode is answering is "How can I run Python programs?" and it jumps straight into "Getting Started with JupyterLab", I wonder whether this could easily be missed.

Would the lesson maintainers be open to a small PR to add a short paragraph at the top of the episode that makes this distinction earlier, or do you think that this would overcomplicate things?

alee:
I think that would be a useful clarification, and would welcome a PR on this

harryjmoss:
Attempted to address this in #551

PR #551, opened 2021-05-13
Explain alternative python script execution methods earlier in lesson

harryjmoss:
As noted in #539 - moves justification for using JupyterLab and an explanation of alternative methods of running python scripts to the start of the lesson

### DIFF
--- a/_episodes/01-run-quit.md
+++ b/_episodes/01-run-quit.md
@@ -21,11 +21,21 @@ keypoints:
 - "Markdown does most of what HTML does."
 ---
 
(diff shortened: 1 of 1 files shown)


alee (approved):
Thanks for the PR! This looks good to me, and helps the content flow better. 

Does this sufficiently address #539 @jatonline ?

jatonline:
This is exactly the sort of clarification I was thinking of! Thanks very much @harryjmoss for your effort in making these changes and @alee for taking the time to review.

PR #553, opened 2021-05-14
Remove matplotlib inline magic

martinosorb:
Running `%matplotlib inline` is no longer necessary when we want to directly display figures in a Jupyter notebook. This has been the default behaviour of Jupyter for several years, since a few versions ago.

There are no mentions of "magic" commands elsewhere in this lesson, so I feel it would be quite confusing if not properly explained, and would be a loss of time if explained in detail (since I would also argue that these later episodes should be more about python than about IPython).


vahtras:
Thanks, your observation is correct. Looks good to me!

alee (approved):
Thanks for catching that!

Issue #554, opened 2021-06-21
Suggestion: introduce getcwd and listdir in first lesson, and remove Markdown


srtee:
I have helped in a few courses run based on the Gapminder material, and in common with Issue #528, getting to the correct directory and accessing the files is a common source of grief. My suggestion would be:

1. put in material in [Episode 1](https://github.com/swcarpentry/python-novice-gapminder/blob/gh-pages/_episodes/01-run-quit.md) covering working directories in Python, and
2. remove the material on Markdown from the same Episode.

The high-level argument for (2) is that Markdown is not mentioned throughout the rest of the course material -- furthermore, it is more a feature of Jupyter and other notebook IDEs rather than Python itself. A Python learner will often be stumped by "File not found"-type errors from not knowing how to navigate directory structures in-script (including in the course itself!), far more than wanting to put a notebook cell into Markdown and not knowing how to.

This would entail the following changes:

**Objectives**

- ~Create Markdown cells in a notebook.~
- Print the current working directory and subdirectories and list their contents

**Key Points**

- ~The Notebook will turn Markdown into pretty-printed documentation.~
- ~Markdown does most of what HTML does.~
- A Python script / JupyterLab notebook has a working directory.
- The script / notebook will look for files in the working directory using a relative path.

and replacing the Markdown content with some exposition around an exercise to open a new Python cell, run the following commands, and describe the results:

`from os import getcwd, listdir`
`print(getcwd())`
`print(getcwd()+'/data')`

`print(listdir('data'))`
`print(listdir(getcwd() + '/data'))`

Learners can be reassured that they will understand the syntax a lot more later on in the course, but for now, they are just learning how to check their working directory and make sure they have their Gapminder data files. Any errors can be handled by instructors and helpers on the spot, making it much more likely that learners will have their scripts / notebooks open in the correct directories and that their data files are already accessible for future use. By moving out the Markdown material, this change will not lengthen the course (and indeed might shorten it by frontloading all the environment setup).

srtee:
Alternatively, the material on getcwd and listdir could be placed at the end of [Episode 6 (Libraries)](https://github.com/swcarpentry/python-novice-gapminder/blob/gh-pages/_episodes/06-libraries.md), which will showcase a very practical and essential example of Python library importing.

[...]

PR #559, opened 2021-08-06
Intro to os module

izaakm:
Addresses issue #554.

Changes:

- Remove markdown content from Episode 1.
- Add content about the `os` library to Episode 6, specifically, introduce the `getcwd`, `listdir`, and `chdir` methods.


alee:
Thanks for the PR! I'm inclined to move the markdown removal into a separate PR so others in the community can chime in on whether they think it's worth removing. 

We could also mark it as optional so instructors know to assess their learners when they are teaching and can skip over it if most people know / understand how it works.

Thoughts from others?

martinosorb:
I'd just like to point out that Pathlib is a new library part of Python's core libraries, and is now considered a much better standard for filesystem operations. `Path.cwd()`, `Path("...").iterdir()` can do the same as `os.getcwd()` and `os.listdir()`. As for `chdir`, I personally don't use it and think changing wd tends to cause trouble.

Issue #460, opened 2020-01-13
"Variables and assignments" - lesson and exercises improvements


mforneris:
## General feedback

I had the chance to teach Episodes 1 to 7 to a class of researchers in molecular biology with little experience in programming. The class really liked the course, the pacing and the amount of material that was covered. It was also very nice to go into practical applications early on to keep the students engaged.

#### Course timing
It actually took us 2 days (from 9:30 to 5) to cover the whole material, only adding an episode about numpy. I think the timing reported within the episodes are too short for a novice audience (especially if instructors take some time to correct a handful of exercises after giving some time to solve them).

#### Comments should be introduced here

The concept of comments within python scripts has not been introduced in this episode but it will be only in Episode 4. I suggest to introduce it here since it is widely used in the exercises and people had a hard time interpreting it.

## Variables and assignments Exercises 

Students found some of the exercises in "Variables and assignments" (http://swcarpentry.github.io/python-novice-gapminder/02-variables/) challenging to understand. I'll summarize here the comments from the audience and some suggestions on how to improve the exercises.

#### Swapping Values

The audience was just introduced with the concept that code gets executed line by line. In addition, comments are necessary to understand what is written here. I would suggest restructuring the exercise in this way:  

> `# Consider the code below. What is the value of each variable after each line of code gets executed? Fill in the table with the value of the variables in this program after each statement is executed`
>  x = 1.0    
y = 3.0    
swap = x  
x = y      
y = swap  
> `# Line  # Value of x   # Value of y   # Value of swap `
> `# Line 1 #             #             #             #             `
> `# Line 2 #             #             #             #             `
> `# Line 3 #             #             #             #             `
> `# Line 4 #             #             #             #             `
> `# Line 5 #             #             #             #             `

The students also found a bit challenging to understand the meaning of the exercise (swapping value between variables) at this stage and as a first exercise in the course. I think it would be nice to move it to the end as the last exercise or before "Slicing concepts" since it wraps up well many concepts introduced in this lesson.

#### Slicing concepts

This exercise was again a bit hard to understand by the students. I would reformulate the question so students should write some code instead of reasoning on what is written there. Many students were asking what is "low" or "high" or if they were variables. For example, I would rephrase the exercise in this way:

> `# Given the string:`
> full_name = "James Bond"
> `# What would these lines return?`
> `# 1. full_name[7:10]`
> `# 2. full_name[6:]`
> `# 3. full_name[:4`
[...]

PR #567, opened 2021-09-08
Update Slicing Concepts exercise (part of issue #460)

betsybookwyrm:
Updates the Slicing Concepts exercise of the Variables lesson to:
- use concrete examples
- use the standard term 'stop' for the end of a range in Python
- add an example of using two negative indices

---

Addresses the Slicing Concepts part of issue #460 , but not the other ideas in there. Follows the model given by @mforneris, and also adds the double negative indices example suggested by @brianmatthewthomas .

---

I haven't managed to get `make lesson-check` work for me (I assume that's the best way for me to test my changes don't break the page formatting locally before submitting?) so this is unchecked, sorry.

Also, sorry about the whitespace changes, I'm having trouble committing the file without making those changes for some reason.

If either of these are problems, let me know and I'll put more time into fixing them!


_episodes/02-variables.md
```diff
@@ -334,20 +334,31 @@ print(ewr_422_yY, 'is', flabadab, 'years old')
...
 >
 > > ## Solutions
 > >
-> > 1. `thing[low:high]` returns a slice from `low` to the value before `high`
-> > 2. `thing[low:]` returns a slice from `low` all the way to the end of `thing`
-> > 3. `thing[:high]` returns a slice from the beginning of `thing` to the value before `high`
-> > 4. `thing[:]` returns all of `thing`
-> > 5. `thing[number:some-negative-number]` returns a slice from `number` to `some-negative-number` values from the end of `thing`
-> > 6. If a part of the slice is out of range, the operation does not fail. `atom_name[0:15]` gives the same result as `atom_name[0:]`.
+> > 1.  `species_name[3:8]` returns the substring `'acia b'`
```
alee, line 356: minor typo, this should be `[2:8]`

_episodes/02-variables.md
```diff
@@ -334,20 +334,31 @@ print(ewr_422_yY, 'is', flabadab, 'years old')
...
+> > 2.  `species_name[11:]` returns the substring `'folia'`, from position 11 until the end
+> > 3.  `species_name[:4]` returns the substring `'Acac'`, from the start up to but not including position 4
+> > 4.  `species_name[:]` returns the entire string `'Acacia buxifolia'`
+> > 5.  `species_name[11:-3]` returns the substring `'fo'`, from the 11th position to the third last position
+> > 6.  `species_name[-5:-3]` also returns the substring `'fo'`, from the fifth last position to the third last
+> > 7.  If a part of the slice is out of range, the operation does not fail. `species_name[0:20]` gives the same result as `species_name[0:]`
```
alee, line 362: maybe add an `and species_name[:20]` to this line?

alee (changes requested):
Thank you for the PR! I like the concrete example, and think this is a solid improvement to the challenge. 

I'd never tried `make lesson-check` until now :sweat_smile:  but it does look like it failing due to a missing ruby dependency that wasn't installed on my local system. 

If you've got Docker installed `make docker-serve` should work to render the lesson and bring up a local webserver that lets you see if the formatting still looks OK at localhost:4000

Thanks again!

_episodes/02-variables.md
betsybookwyrm: Thanks for catching that!

betsybookwyrm:
I've incorporated the typo fix and the suggestion above :) Also rebased on gh-pages branch.

Also, thanks @alee for the tip about `make docker-serve`, I got that working, it's very nice!

PR #569, opened 2021-09-10
20210909 styles upgrade

fmichonneau:
it looks like the template for this lesson hasn't been updated in a while and it contains an older version of jQuery that is known to have some security vulnerabilities. This PR brings the latest version of the lesson template.


souravsingh:
LGTM

PR #570, opened 2021-09-11
Ep02 minor fixes/typos

vinisalazar:
Hi,

Here are some minor fixes. If merged, this PR will:

- Close slice with bracket rather than parenthesis in descriptive text
- Add print statement to solution of 'Predicting Values' challenge
- Fix punctuation of 'Slicing Concepts' challenge

Best,
Vini

### DIFF
--- a/_episodes/02-variables.md
+++ b/_episodes/02-variables.md
@@ -153,7 +153,7 @@ h
 *   We take a slice by using `[start:stop]`, where `start` is replaced with the
     index of the first element we want and `stop` is replaced with the index of
     the element just after the last element we want.
-*   Mathematically, you might say that a slice selects `[start:stop)`.
+*   Mathematically, you might say that a slice selects `[start:stop]`.
 *   The difference between `stop` and `start` is the slice's length.
 *   Taking a slice does not change the contents of the original string. Instead,
     the slice is a copy of part of the original string.
--- a/_episodes/02-variables.md
+++ b/_episodes/02-variables.md
@@ -342,15 +342,15 @@ print(ewr_422_yY, 'is', flabadab, 'years old')
 > 3.  What does `thing[:high]` (without a value before the colon) do?
 > 4.  What does `thing[:]` (just a colon) do?
 > 5.  What does `thing[number:some-negative-number]` do?
-> 6.  What happens when you choose a `high` value which is out of range? (i.e., try `atom_name[0:15]`) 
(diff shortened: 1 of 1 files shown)


_episodes/02-variables.md
```diff
@@ -153,7 +153,7 @@ h
 *   We take a slice by using `[start:stop]`, where `start` is replaced with the
     index of the first element we want and `stop` is replaced with the index of
     the element just after the last element we want.
-*   Mathematically, you might say that a slice selects `[start:stop)`.
+*   Mathematically, you might say that a slice selects `[start:stop]`.
```
alee, line 156: The `[start:stop)` expression (which is invalid Python syntax) refers to [mathematical interval notation](https://en.wikipedia.org/wiki/Interval_(mathematics)#Notations_for_intervals) and means "include the value at `start`, but do not include the value at `stop`". If this is confusing to others however I'm open to adding a footnote or link or finding a different way to explain it.

_episodes/02-variables.md
```diff
@@ -339,15 +342,15 @@ print(ewr_422_yY, 'is', flabadab, 'years old')
 > 3.  What does `thing[:high]` (without a value before the colon) do?
```
alee, line 342: This PR will have some conflicts with https://github.com/swcarpentry/python-novice-gapminder/pull/567/files which I'd like to merge in as well - perhaps we can revisit it after that is merged?

alee (changes requested):
Thanks for the PR! I would like to get https://github.com/swcarpentry/python-novice-gapminder/issues/460 merged in first and discuss the issues with the slice interval notation

alex-ball:
According to the Wikipedia link (and this is indeed how I remember it) mathematical interval notation uses a comma rather than a colon for separating the limits. To make it clearer that this isn't Python, perhaps it would be better to have `[start, stop)` or, even better, avoid the monospace and write it [*start*, *stop*).

vahtras:
We also have that a mathematical interval [x, y) is more like a set of
numbers and fundamentally different from a Python sequence of numbers where
there is an implied order.

alee:
> We also have that a mathematical interval [x, y) is more like a set of numbers and fundamentally different from a Python sequence of numbers where there is an implied order.

Based on this discussion I'm inclined to simply remove the entire statement as it seems to be muddying the waters more than clarifying anything 😅 

thanks for the feedback!

alee (approved):

PR #572, opened 2021-09-14
Adds links for Jupyter Notebook and JupyterLab

sborrego:
I added the website links but since there are so many instances of missing website links I wasn't sure if this was intentionally left the way it was.

### DIFF
--- a/_episodes/01-run-quit.md
+++ b/_episodes/01-run-quit.md
@@ -22,7 +22,7 @@ keypoints:
 ---
 
 Many software developers will often use an integrated development environment (IDE) or a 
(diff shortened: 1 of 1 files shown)


alee:
I think this is actually an issue with some of our local changes to https://github.com/swcarpentry/python-novice-gapminder/blob/gh-pages/_includes/links.md getting removed by the recent styles merge - we can fix these missing links by adding e.g., a [jupyterlab] entry to the links.md file instead

Thanks for catching this!

alee:
45aa49de36ba2a0a3700c5fb6240a1c553e36f48 has been pushed to fix this link - thanks again @sborrego!

PR #577, opened 2021-12-23
Clarify explanation of "How Large is a Slice?"

alex-ball:
The solution for "How Large is a Slice?" says the the expression `values[low:high]`, or possibly `high - low`, will only work if `high` is less than the total length of the list `values`. This isn't quite correct on either interpretation.

As shown in Episode 2, `values[low:high]` will always return a list, even if one or both of the limits are out of bounds. The length of the slice will be `high - low` if and only if `high` is equal to or less than the length of the list; it will be zero if both `high` and `low` are greater than the length of the list, and somewhere in between otherwise.

This pull request suggests alternative wording that I think would be clearer and more accurate.

### DIFF
--- a/_episodes/11-lists.md
+++ b/_episodes/11-lists.md
@@ -236,14 +236,14 @@ IndexError: string index out of range
 
 > ## How Large is a Slice?
 >
-> If `low` and `high` are both non-negative integers,
-> how long is the list `values[low:high]`?
+> If `start` and `stop` are both non-negative integers,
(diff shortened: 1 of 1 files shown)


alee (commented):
thanks for the PR! This is a good clarification to make. 

Should we replace usage of `low` and `high` variables here with `start` and `stop` as well to match up with episode 2?

alex-ball:
Yes, good idea. It would certainly help to make the link more obvious.

PR #579, opened 2021-12-23
Avoid setting encapsulation challenge too early

alex-ball:
In episode 16, the panel "Encapsulation of an If/Print Block" asks the learner to turn the if-block into a function a total of three times:

- "Please re-write the code so that the if-block is folded into a function."
- "What function definition will make it functional?"
- "Create a function definition..."

Only the last of these is needed. The first one is particularly unhelpful as it comes before the task is properly defined.

This pull request removes the duplication, and instead provides a potential rationale for wanting to do the task in the first place. (It also removes some trailing spaces.)


alee (approved):
Thanks for catching this!

PR #583, opened 2022-01-04
Improve the solution to "Initializing" (Ep. 13)

alex-ball:
This PR seeks to improve three aspects of the solution to the "Initializing" challenge in episode 13.

 1. The solution includes the expression `== None`, which is contrary to PEP8 and thus widely frowned on. This PR changes it to `is None`, but because no-one would expect a learner to know about this, it includes a note of explanation.
 2. I find the statement "the `values` list is iterated three times" rather confusing, since `values` is only iterated once: the inner loops iterate over exactly two values each, not the whole list. (Each member of `values` appears in three loops, but that is not the same thing as the entire list being iterated three times.) I have suggested alternative wording.
 3. I have performed some tests and found that the second solution is not, in fact, the most efficient, if by that we mean the quickest in general.

By way of comparison, I timed the three solutions in this PR running over a shuffled list of 100_000 integers, 1000 times each, and on my machine get these timings:

- First solution (three loops): ~ 19s.
- Second solution (one loop with four comparisons): ~ 5.5s.
- Just using min() and max() (two loops, one comparison each): ~2.1s.

For completeness, I also tested a solution that did away with testing for None (so one loop with two-ish comparisons):

```python
smallest, largest = values[0], values[0]
for v in values:
    if v < smallest:
        smallest = v
    elif v > largest:
        largest = v
print(smallest, largest)
```

and that took ~3.8s.

### DIFF
--- a/_episodes/13-conditionals.md
+++ b/_episodes/13-conditionals.md
@@ -385,9 +385,9 @@ final velocity: 30.0
 > > values = [-2,1,65,78,-54,-24,100]
 > > smallest, largest = None, None
 > > for v in values:
-> >     if smallest == None or v < smallest:
+> >     if smallest is None or v < smallest:
 > >         smallest = v
-> >     if largest == None or v > largest:
+> >     if largest is None or v > largest:
 > >         largest = v
 > > print(smallest, largest)
 > > ~~~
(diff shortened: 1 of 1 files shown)


alee (approved):
Thanks for the PR! Great analysis + explanation and apologies for the long delay in review; Omicron and other real life concerns took over most of this month.

We could also analyze the [generated byte code](https://docs.python.org/3/library/dis.html) to see why the min/max methods are an improvement over the single loop but that seems overkill🤓

PR #584, opened 2022-01-04
Provide link to documentation of float('Inf')

alex-ball:
The solution to "Minimum File Size" in Episode 14 includes a use of `float('Inf')`; this is not introduced previously and might well come as a surprise.

For the benefit of those learners who went for `1e10` or similar, this PR cushions the blow a bit by explaining that, while using a big enough number is valid, using infinity is better. It then provides a link to the documentation that explains about this and other special strings that can be passed to `float()`.


alee (approved):
Thanks, this is helpful context!

PR #585, opened 2022-01-04
Correct solutions for "Simulating a dynamical system" (Ep. 16)

alex-ball:
The challenge "Simulating a dynamical system" in Episode 16 asks for values at times `[0,1,...,t_final]` but the solutions provide values at times `[0,1,...,t_final - 1]`.

For example, in part two of the solution, `population` is initialized with the `initial_population`, and then the loop runs 9 times (`range(1, 10)`) to give a total of ten values, the last of which is the value at *t* = 9. A tenth run is needed to get what was asked for, so the loop should be:

```python
for t in range(t_final):
    population.append( logistic_map(population[t], r) )
```

Happily, this looks even more like the function in part one, where you get the value at *t* + 1 in terms of the value at *t*, rather than the value at *t* in terms of the value at *t* - 1.

This PR supplies the corrected code for parts two and three, and the value of the population at *t* = 10, 100, 1000 instead of at *t* = 9, 99, 999.

### DIFF
--- a/_episodes/16-writing-functions.md
+++ b/_episodes/16-writing-functions.md
@@ -529,8 +529,8 @@ result of call is: None
 > 2. Using a `for` or `while` loop, iterate the `logistic_map` function defined in part 1, starting
 >    from an initial population of 0.5, for a period of time `t_final = 10`. Store the intermediate
 >    results in a list so that after the loop terminates you have accumulated a sequence of values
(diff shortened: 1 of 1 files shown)


alee (approved):
I like this change as it simplifies the code and removes the off-by-one computations. If we were only interested in `t_0 .. t_final-1` we could also iterate over `range(t_final - 1)` instead of subtracting one from the index every time while iterating over `1 .. t_final`.

Thanks for the PR!

PR #586, opened 2022-01-17
Added `random.choice` as sampling option (581)

### LINKED ISSUE #581: Selecting random character from string (Ep. 6) missing a good solution
The challenge "[Locating the right module](http://swcarpentry.github.io/python-novice-gapminder/06-libraries/index.html#locating-the-right-module)" has several solutions, but I think the most obvious one is missing.

The solutions given are:

```python
print(bases[random.randrange(len(bases))])
print(bases[random.randint(0, len(bases) - 1)])  # implied
print(random.sample(bases, 1)[0])
```

with the postscript "\[There are] also other functions you could use, but with more convoluted code as a result." What about this?

```python
print(random.choice(bases))
```

This is, I would argue, the most keystroke-efficient solution and most intuitive to read; it certainly does not fall into the "more convoluted" category. Is the reason for missing it out because it answers exactly the use case in the challenge, and is therefore not applicable to adjacent challenges (e.g. picking more than one distinct character; doing something else with the index)?

guydav:
Also removed the reference to being able to find more convoluted ways to solve the problem, as suggested in the issue discussion.

Resolves #581 .

### DIFF
--- a/_episodes/06-libraries.md
+++ b/_episodes/06-libraries.md
@@ -164,12 +164,12 @@ cos(pi) is -1.0
 >
 > > ## Solution
 > >
-> > The [random module][randommod] seems like it could help you.
+> > The [random module][randommod] seems like it could help.
 > >
 > > The string has 11 characters, each having a positional index from 0 to 10.
-> > You could use either `random.randrange` or `random.randint` functions
-> > to get a random integer between 0 and
-> > 10, and then pick out the character at that position:
+> > You could use the [`random.randrange`](https://docs.python.org/3/library/random.html#random.randrange)
+> > or [`random.randint`](https://docs.python.org/3/library/random.html#random.randint) functions
+> > to get a random integer between 0 and 10, and then select the `bases` character at that index:
 > >
 > > ~~~
 > > from random import randrange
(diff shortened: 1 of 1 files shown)


alee:
Hi @guydav - this LGTM and I'm happy to merge, but for posterity you should put these changes on a branch in your fork to avoid future issues with keeping your fork and this upstream repository synced. You could always delete and re-fork but it's a good habit to always submit pull requests to upstream from a new branch in your fork.

guydav:
Hi @alee -- I don't know how I forgot to do that, my apologies. Would you prefer if resubmit the PR in a separate branch?

alee:
If you have the time and would like the GitHub practice that's fine, otherwise just wanted you to be forewarned that after merging your fork's gh-pages and this upstream will be out of sync and you'll likely have to do some Git surgery to get things lined up again

alee:
Great! You should be able to do it straight from the GitHub web interface by clicking on the branch dropdown at the top left of the Code tab:

image

As soon as the branch is created you can submit the PR from it. Don't worry about bringing the branch up to date with upstream, I don't think there are any conflicts. 

After that's done you'll still need to reset your fork's gh-pages to bring it in line with upstream though..

guydav:
Done. See #592

PR #589, opened 2022-02-08
Remove unexplained augmented assignment operators from examples

### LINKED ISSUE #582: Augmented assignment operator introduced without comment
The solution to the "[Cumulative Sum](http://swcarpentry.github.io/python-novice-gapminder/12-for-loops/index.html#cumulative-sum)" challenge in Episode 12 uses the augmented assignment operator `+=` without comment or explanation. In the rest of the episode, the more verbose `total = total + whatever` syntax is used. The only other occurrence of `+=` is in Episode 18.

My feeling is that `+=` is a useful shorthand but – given how similar it looks to comparisons like `!=`, `>=`, `<=` – deserves a little explanation for anyone who hasn't seen it before. How about having a callout just before "Cumulative Sum" that explains it, then converting "Identifying Variable Name Errors" to use it as well? That would help with retaining it for Episode 18.

Another solution would be to replace the instances in both episodes with the explicit, longer version, which would be less cognitive load but perhaps a bit of a shame.

naclomi:
Resolves #582 per issue discussion.

I did leave _one_ instance of `+=` in the exercise in episode 18, since its point was to emphasize the value of readable rather than terse code and I feel that running into unfamiliar syntax in the wild is a pretty common experience with codebases that need refactoring. Happy to remove it for overall consistency, though.

### DIFF
--- a/_episodes/12-for-loops.md
+++ b/_episodes/12-for-loops.md
@@ -322,7 +322,7 @@ print(total)
 > cumulative.append(total)
 > for number in data:
 > cumulative = []
-> total += number
+> total = total + number
 > total = 0
 > print(cumulative)
 > data = [1,2,2,5]
@@ -334,7 +334,7 @@ print(total)
 > > data = [1,2,2,5]
 > > cumulative = []
 > > for number in data:
-> >     total += number
+> >     total = total + number
 > >     cumulative.append(total)
 > > print(cumulative)
 > > ~~~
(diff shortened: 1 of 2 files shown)


vahtras:
The changes are ok, but you should really have a dedicated development branch (not your gh-pages) with a descriptive name for which the pull request is done. Otherwise, if you later push an unrelated change to your gh-pages branch it will reopen this PR.

naclomi:
See PR #590

PR #595, opened 2022-03-04
Fix JupyterLab link in index.md (#594)

jeskowagner:
Repeated link to jupyter.org for JupyterLab in links.md to address #594. Sorry for the messy PR with three commits when one would've sufficed (not sure how to clean it up).

Issue #594, opened 2022-03-02
JupyterLab link broken on index.md


sborrego:
I just wanted to point out that the link for JupyterLab is broken on **line 10** in [index.md](https://github.com/swcarpentry/python-novice-gapminder/edit/gh-pages/index.md). It does not render correctly on the [webpage](http://swcarpentry.github.io/python-novice-gapminder/).

Hope this helps.

alee:
Thanks for catching this @sborrego !

PR #596, opened 2022-03-06
Fix JupyterLab link in index.md

jeskowagner:
Repeated link to jupyter.org for JupyterLab in links.md to address #594. This PR is a copy of #595, but from a new branch.


alee (approved):
Thanks for the fix!

Issue #598, opened 2022-03-31
Adding caveats for DataFrame.iloc under Pandas Dataframes


candemircan:
Hi!

For the [Pandas Dataframes](http://swcarpentry.github.io/python-novice-gapminder/08-data-frames/index.html) episode, under the `DataFrame.iloc[..., ...]` section, it might be worth mentioning the caveats of this method, i.e. if you add new columns to your data later, an index based selection (as opposed to using column names) can lead to problems. If this is worth adding, I would be happy to make the edit and make a pull request.

Thanks,

Can

richiehodel:
I think a 'caveats' section would be a great addition, @candemircan. The lesson workflow does a good job of introducing learners to using different approaches to slice data frames. However, I wonder if learners may run into trouble when applying some of the knowledge from this lesson (e.g., adding columns to data later and running into index-based selection problems, as you mention). Perhaps immediately after the section "Result of slicing can be used in further operations", a section could be added to demonstrate the caveats and how learners might run into trouble. After completing the lesson, if learners start adding on further operations to .iloc, they might run into a "SettingWithCopyWarning", and be unsure why it is happening. Maybe addressing this specific warning is beyond the scope of the lesson, but including a brief section demonstrating the caveats would be valuable.

chillenzer:
Hi there,

I agree and I would be happy to write something up in that direction. However, I would discuss this further before starting an attempt:
Before getting to the point, I want to suggest going even further and teaching the `.loc` method before `.iloc` because it is more `pandas`othic (in the sense of `python`ic): If you reach for `pandas` instead of `numpy` that should be because there is a special meaning to rows and columns of your data (and not their indices) and, if so, you should assign meaningful labels and use those.

Anyways, both methods have their caveats. A list from the top of my head would be (inlcuding yours):

1. _Choose `.loc` over `.iloc` if possible_ because operations might change the index in a non-obvious way (and above).
2. _Refrain from combining more than one `.loc` and/or `.iloc`_ due to the mentioned `SettingWithCopyWarning`. I think that should be best practice even if not setting anything because you might later copy that expression for setting the same elements. Also, try to never return `.(i)loc[...]` from a function because there is no guarantee what will happen outside the function (another `.(i)loc` perhaps?).
[...]

PR #602, opened 2022-07-15
minor changes to 08-data-frames

paulmaxus:
Commit 1 adds print statements on 2 lines of code:
To be consistent with the rest of the episode's code, the dataframes in question should be printed explicitly.
Leaving out the print statement suggests that this would also work outside of the jupyter environment. 

Commit 2 changes the explanation of the .loc method:
1. dictionaries were not explained yet in this lesson; I deleted that part
2. location can be specified by column names as well; I've added that information


alee (approved):
Thanks for the PR! I think these are good incremental improvements though there are further refinements expected to this section from #598

PR #608, opened 2022-08-10
Update 02-variables.md

potus28:
Made a small change according to Issue 599. Bullet points for variable names being case sensitive have been moved to the front, and an additional bullet point was created for making variable names meaningful.


alee (approved):
LGTM, thoughts from others?

PR #609, opened 2022-08-11
Remove premature function example

ndporter:
Removes the example of a function in pandas that is given here before functions are taught (episode 16). At current, it would be confusing at best unless people happen to know `def` already.

This example could also be moved to the functions lesson instead, although applies are a fairly complex topic to introduce only in passing. They are quite valuable but should be given a more full discussion if we use them.


vahtras:
Makes sense to me / Olav

alee (approved):
Thanks for catching this and submitting the PR! I moved the function from ep 13 -> 16 for now, just to keep it as an option for instructors but you also make good points as to completely removing or refining it. That might be better served as a separate issue / PR however.

PR #610, opened 2022-08-18
Update 08-data-frames.md

chaszg:
There was a missing `#` on line 554. I added one so that the markdown render doesn't recognize the comment as Python code.


alee (approved):
Thanks for catching this!

PR #623, opened 2023-01-13
replace `str.strip()` with `str.replace()`

n2ygk:
`str.strip()` actually removes each character in the character class string. So `.strip('gpdaPerc_')` would also do the same thing and is confusing.  I made a live typo during the workshop and a student asked me how was it that it still worked when I had typed `gdp_Percap_` even though there's no `_` between `gdp` and `Per`!

Reference: https://pandas.pydata.org/docs/reference/api/pandas.Series.str.strip.html and https://pandas.pydata.org/docs/reference/api/pandas.Series.str.replace.html

### DIFF
--- a/_episodes/09-plotting.md
+++ b/_episodes/09-plotting.md
@@ -71,10 +71,10 @@ data = pd.read_csv('data/gapminder_gdp_oceania.csv', index_col='country')
 # Extract year from last 4 characters of each column name
 # The current column names are structured as 'gdpPercap_(year)', 
 # so we want to keep the (year) part only for clarity when plotting GDP vs. years
-# To do this we use strip(), which removes from the string the characters stated in the argument
-# This method works on strings, so we call str before strip()
+# To do this we use replace(), which removes from the string the characters stated in the argument
+# This method works on strings, so we call str before replace()
 
-years = data.columns.str.strip('gdpPercap_')
+years = data.columns.str.replace('gdpPercap_', '')
 
 # Convert year values to integers, saving results back to dataframe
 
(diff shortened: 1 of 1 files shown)


vahtras:
Thanks that makes sense to me

alee (changes requested):
thanks for catching this! Looking more closely at this section makes me realize there are a few more areas that could use improvement:

- would be nice to embed a link to the https://pandas.pydata.org/docs/reference/api/pandas.Series.str.replace.html in the call
- I think the text "call str before replace()" beforehand isn't quite correct or at least is a bit misleading as we aren't calling `str` but instead accessing the `str` attribute of the Series class (e.g., https://pandas.pydata.org/docs/user_guide/text.html#string-methods)

I'm happy to add the changes in but if you have time to also include them it would be much appreciated!

n2ygk:
> thanks for catching this! Looking more closely at this section makes me realize there are a few more areas that could use improvement:
> 
> * would be nice to embed a link to the https://pandas.pydata.org/docs/reference/api/pandas.Series.str.replace.html in the call
> * I think the text "call str before replace()" beforehand isn't quite correct or at least is a bit misleading as we aren't calling `str` but instead accessing the `str` attribute of the Series class (e.g., https://pandas.pydata.org/docs/user_guide/text.html#string-methods)
> 
> I'm happy to add the changes in but if you have time to also include them it would be much appreciated!

@alee so this fragment is in a python code comment (not markdown text). I guess I'll add the links to the bullet list above the code.

alee (approved):
LGTM, thanks!

PR #625, opened 2023-04-06
refactor: restructure episode 3

### LINKED ISSUE #600: Restructure the episode 3 - Data Types and Type Conversion
It would be more clear to move "Types control what operations (or methods) can be performed on a given value." and "Must convert numbers to strings or vice versa when operating on them." together and move them to the end of the episode. 
The structure of the 3rd episode will be talking about different data types first, operations of same type of data, then the operations between different data types (floats and integers can be mixed, but string and number couldn't be mixed,) 
Hope it makes sense.

yld-weng:
Move sections `Types control what operations (or methods) can be performed on a given value` and `Must convert numbers to strings or vice versa when operating on them.` together as they are both focus on operations between different types.

Close #600.

### DIFF
--- a/_episodes/03-types-conversion.md
+++ b/_episodes/03-types-conversion.md
@@ -12,12 +12,12 @@ objectives:
 keypoints:
 - "Every value has a type."
 - "Use the built-in function `type` to find the type of a value."
-- "Types control what operations can be done on values."
 - "Strings can be added and multiplied."
 - "Strings have a length (but numbers don't)."
-- "Must convert numbers to strings or vice versa when operating on them."
 - "Can mix integers and floats freely in operations."
 - "Variables only change value when something is assigned to them."
+- "Types control what operations can be done on values."
+- "Must convert numbers to strings or vice versa when operating on them."
 ---
 ## Every value has a type.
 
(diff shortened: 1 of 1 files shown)


alee:
Thank you for the PR! Though I'm sorry to say that after taking a closer look at the episode with the rearranged key points / sections I'm not convinced this is actually an improvement, despite encouraging people to take a closer look in #600 

For clarity, this PR and changeset makes the sections and corresponding key points go follow this thread:

- Every value has a type.
- Use the built-in function type to find the type of a value.
- Strings can be added and multiplied.
- Strings have a length (but numbers don’t).
- Can mix integers and floats freely in operations.
- Variables only change value when something is assigned to them.
- **Types control what operations can be done on values.**
- **Must convert numbers to strings or vice versa when operating on them.**

The original:

- Every value has a type.
- Use the built-in function type to find the type of a value.
- **Types control what operations can be done on values.** *then we follow up with the kinds of operations types enable*
- Strings can be added and multiplied.
- Strings have a length (but numbers don’t).
- **Must convert numbers to strings or vice versa when operating on them.** *start to delve into type coercion*
- Can mix integers and floats freely in operations.  _continue type coercion_
- Variables only change value when something is assigned to them. _this does seem a bit out of the blue and starts to scratch the surface of references, mutation, and variable assignment semantics but also seems an important concept to convey to learners_

I feel like the original's ordering works fine but am open to thoughts from others who have taken the course or taught the course. Thoughts?

PR #627, opened 2023-04-17
Update to final version of styles lesson template

tobyhodges:
Bringing the lesson template up-to-date in preparation for the planned infrastructure transition.

This is a very large set of changes. You do not need to check through the whole diff, but feel free to pull the branch and build the site locally if you would like to. (It built successfully on my system, and I am 99% sure it will do so on GitHub too when merged.)


alee:
worked on my local system using `make docker-serve` also, thanks @tobyhodges !

Issue #636, opened 2023-05-03
Examples / exercises encourage anti-patterns in 12-for-loops.md


augeas:
There are two examples/exercises whose solutions encourage anti-patterns, miss basic Python features introduced previously, and use the language non-idiomatically. We shouldn't add the confusing cognitive load of "Here's an example, but *don't* do it this way.".

1) Intro to the accumulator pattern:

To be worthwhile, the accumulated values should be *somewhat* non-trivial, students should not leave the session thinking that a for loop is required to merely sum over an iterable:

```
sum_to_10 = sum(range(1, 11))
```
This is a missed opportunity to introduce "sum" in the previous built-in functions section.
Solving the problem of range(10) going from 0-9 would be better solved by explaining that one wanted range(1,11) rather than a superfluous extra addition. 
If a cheap-and-cheerful Monte-Carlo simulation to estimate Pi is too much, maybe just count the number of heads in some coin-tosses?

2) Reversing a String:

This is *not* a good use for a for loop. Slice operators have already been introduced previously, and students should leave thinking that this is how you reverse a string.
```reversed_string = 'This is not a palindrome![::-1]```

Easy. Yes, this demonstrates that iteratively appending items from a collection will cause them to be in reverse order, but why would anyone do that?

alee:
Thanks for the feedback! Minor pedantic note that the simplest way to sum to 10 doesn't need the first parameter, `sum(range(11))` suffices.

The string reverse slice operator requires a clear explanation to remove it from the realm of magic but I think it would be a great addition to cement the conceptual model of how slicing works in Python.

There are a fair number of things missing from this lesson in its original incarnation that I believe were deliberate design decisions and that I've attempted to honor during my time as a maintainer with the awareness that this scope / vision should evolve over time alongside the instructors and learners using the course...

1. list / dict comprehensions
2. lambdas / functions as first class values
3. classes / OOP
4. call-by semantics
5. any others?

martinosorb:
There is also a discussion on whether to move conditionals up before for loops. For-loop exercises with conditionals should make much more sense and be more intuitive, imho.

Issue #548, opened 2021-05-04
Reorganisation of lesson


pelahi:
The current lesson plan seems out of order to me. Based on my experiences teaching programming and python, the first topics should cover more basic concepts and end with using functions and libraries to plot results. I have opened a pull request #547 showing the reorganisation of the lesson.

vinisalazar:
That is a fair point. While I tend to agree that presenting basic syntax and then introducing external libraries and functions seems like a more 'natural' order, I believe that the advantage of the current order of episodes is to quickly expose learners to what can be done with Python (reading tabular data, summarising the data, and plotting). In my experience, this makes learners more interested in the lesson and improves their experience in the later episodes (lists, conditionals, functions). Curious to know what other people think.

vahtras:
I believe that the philosophy of the current order is to "introduce the
useful stuff early" rather than in the end after the basics. It is
intentional, not an oversight.  Although I admit that I had a similar
reaction the first time I saw the material but I do see the value now.
Another point is that in workshops it is not uncommon to run out of time,
so you would rather not have to rush through the good stuff in the end or
even to leave it out.

Den tors 6 maj 2021 14:25Vini Salazar ***@***.***> skrev:

pelahi:
Perhaps I am being a bit harsh but if in practice there is not enough time to cover what would qualify as "basics", then it should be left out and at the outset, the material should make clear that users should have some understanding of some basic Python. I realise there is a lot of material in most carpentries lessons and they are aspirational rather than some stick to. However, having more basic concepts at the end, which in practice is not covered means it should be removed and the lesson restructured so that this knowledge is a prerequisite. 

Even with the philosophy of "introduce useful stuff early", for loops, conditionals and lists are *VERY* useful stuff if you have never been exposed to them. So the organisation does not follow that philosophy anyway. Again, unless you assume most users have some knowledge of programming structures and python, in which case all basic material should be removed. 

Again, I do not want to sound too harsh but a reduction in the number of episodes and a change from large number of episodes to a smaller number of more specific ones would be better. Or it needs to be clear to students that the amount of material is aspirational. I would side on removing unnecessary material. Perhaps the lesson should be reworked to be just Analysis and Plotting with Python. 

Another possibility is to have a non-hands on intro material that highlights the goals of the lesson and then moving on to the more conventional lesson plan but reduce the amount of material in these introductory episodes, perhaps grouping them together. More of a refresher. 

[...]

Issue #113, opened 2016-10-18
Alternative schedule proposal


gvwilson:
Along with a lot of exercises (thanks!), the modified version of this lesson at https://biologyguy.github.io/python-novice-gapminder/ puts episodes in a new order - it takes longer to get to the data analysis payoff, but makes more sense from a building-blocks point of view. What do you think?

andreww:
I can see that the modified order makes more sense, but I suspect it is more important to get to the "payoff" early. I'm also not convinced by separating "Libraries" from the first use of Pandas.

I like to do is split the python material over two days (e.g. afternoon of day 1, and morning of day 2). It would be interesting to know what happens to no-shows on day 2 if this is done and the two orderings are tried for a couple of workshops.

Having said that, I've not actually tried giving this lesson yet (it looks like I will in January) so you may want to take my views with a pinch of salt. Finally, I think this kind of modification for particular workshops is something we should be happy with. Especially if it is in response to something we know about the needs, background or expectation of the attendees at a particular workshop.

pvanheus:
I see Steve (@biologyguy) added an extra hour to the schedule. That's wise, if you can do it. I taught the gapminder lesson at PyConZA recently and ran out of time. If you are going to run out of time, the order that Steve proposes means that you might miss out programming style and you might miss out conditionals even but you'll have covered functions, lists, dataframes and the basics.

bsmith89:
I see three ways of organizing the material with respect to this discussion:

1.  Motivating example (e.g. plotting) first (this is what [swcarpentry/python-novice-inflammation] (//github.com/swcarpentry/python-novice-inflammation) does)
2.  Motivating example early (by the end of the first half-day) ([swcarpentry/python-novice-gapminder](//github.com/swcarpentry/python-novice-gapminder))
3.  Motivating example once learners have all the building blocks ([biologyguy/python-novice-gapminder](//github.com/biologyguy/python-novice-gapminder))

I believe that the best approach of these three will depend on how motivated the learners are.  If everyone's there to learn shell and they've never heard of Python, take approach (1).  If they already _know_ they want to learn Python, (3) avoids confusing/intimidating them early-on.

There's both inter- and intra-workshop heterogeneity in the audience, so I doubt there is one, universally best approach.

lo5an:
What if we add a README notebook to python-novice-gapminder-data.zip that contains an introduction to the GapMinder data, along with some code cells and a plot or two?

[...]

PR #637, opened 2023-05-04
Replace the basic summation example with a simple simulation of coin-…

augeas:
Issue #636:Examples / exercises encourage anti-patterns in 12-for-loops.md 

Replace the example somewhat heavy-handed summation over a loop variable with a simulation of coin tosses:

1) It's *entirely* valid for an accumulator to *not* depend on the loop variable, c.f. Monte-Carlo integration.
2) We shouldn't solve the problem of `range` starting from zero with spurious additions in a loop, just use `range` properly. 3) A trivial summation over the loop variable should be done with the built-in `sum` function.
4) The later exercise with a cumulative sum is a far better example.

Imports have been introduced previously, so it's probably okay to use random.random()
Other examples in the episode use if statements, but this is just more justification to switch the conditionals and loops episodes; conditionals are more fundamental than loops, if you don't have conditional branches, you're not Turing-Complete, and you don't really have a language.

Delete the string revering exercise.
1) Strings, and other sequences are more idiomatically reversed with slice operators.
2) The other exercises with string accumulators are much better.

 On branch loops_grg
	modified:   episodes/12-for-loops.md

### DIFF
--- a/episodes/12-for-loops.md
+++ b/episodes/12-for-loops.md
@@ -147,25 +147,27 @@ a range is not a list: range(0, 3)
 
 - A common pattern in programs is to:
   1. Initialize an *accumulator* variable to zero, the empty string, or the empty list.
-  2. Update the variable with values from a collection.
+  2. Update the variable incrementally, this might use the values of the loop variable, but doesn't have to.
 
 ```python
-# Sum the first 10 integers.
-total = 0
-for number in range(10):
-   total = total + (number + 1)
-print(total)
+# Simulating 100 coin flips.
+import random
+
+total_flips = 100
+total_heads = 0
+for flip in range(total_flips):
+    if random.random() < 0.5:
+      total_heads += 1
+
+print(f'Of {total_flips} coin flips, {total_heads} were heads.')
 ```
 
 ```output
-55
+Of 100 coin flips, 48 were heads.
 ```
 
-- Read `total = total + (number + 1)` as:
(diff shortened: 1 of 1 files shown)


alee:
Thanks for the PR! I like this change personally but would like to gather feedback from the rest of the community as well before moving forward.

I believe the original intent of re-implementing `sum` was to show a novice with minimal programming languages experience how something like creating a summation from 0-N could be implemented. That said, if it were to remain as is I definitely agree a followup with the built-in `sum` as the clearly better solution should be included.

I don't think we introduce format strings in the lesson so there should be some explanation of this either now or earlier.

Please keep in mind that this is not a foundational computer science course and so statements of Turing-completeness as to whether conditionals should be introduced before loops may not track with the intended audience. There have been multiple suggestions to reorder the episodes and ordering and I don't know that we've ever found true consensus yet...

e.g.,

https://github.com/swcarpentry/python-novice-gapminder/pull/294
https://github.com/swcarpentry/python-novice-gapminder/issues/113
https://github.com/swcarpentry/python-novice-gapminder/issues/548

(etc)

martinosorb:
I generally agree on that exercise being not great. I taught it before and left some students puzzled, even though the majority are ok with it. However I would like to point out that your example has a `random.random() < 0.5` line that is not necessarily intuitive. I imagine if I taught that, I would have to stop and explain for quite a while what that is for. Perhaps `np.random.choice((0, 1))` could be an option?

PR #642, opened 2023-05-11
Instructor notes for the split-apply-combine exercise.

PaulineLawrey:
This code snippet often causes a lot of confusion for the participants.

### DIFF
--- a/episodes/08-data-frames.md
+++ b/episodes/08-data-frames.md
@@ -234,6 +234,16 @@ max      13450.401510    16361.876470    18965.055510
 
 ## Group By: split-apply-combine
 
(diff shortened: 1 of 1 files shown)


alee (commented):
Thanks for the PR! I think this is one of the episodes that needs the most additional help, both in the lesson and the instructor guides. I'm happy to merge these changes as there is nothing here yet but would also appreciate additional insight from the community as to how to clearly describe what's going on in this short section:

```python
mask_higher = data > data.mean()
wealth_score = mask_higher.aggregate('sum', axis=1) / len(data.columns)
print(wealth_score)
```

https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.mask.html

PR #643, opened 2023-05-12
Clarify kernel name.

mkcor:
Dear maintainers,

This is minor, but I found myself hesitating for a second while teaching this [yesterday](https://github.com/mkcor/python-prog/blob/33bbbf2b3f3c3999d5e5006dfb59772a398cf50f/_episodes/00-run-quit.md).

I suggest this slight change:

* because the icon on the JupyterLab landing page reads "Python 3 (ipykernel);"
* based on "The IPython kernel is the Python execution backend for Jupyter" from the [official docs](https://ipython.readthedocs.io/en/stable/install/kernel_install.html).

/cc @JohanMabille


martinosorb:
Dear @mkcor I believe 'interactively' is the main point of that sentence. The mention 'in Jupyter' on the other hand is a little redundant, because the sentence before is introducing what happens when a Jupyter notebook is opened... I personally would leave it as it is, or if you have other ideas let us know.

PR #645, opened 2023-05-18
Update 01-run-quit.md

kris-joseph:
Corrected a link to Anaconda Navigator documentation that was resulting in a 404 error.


alee (approved):
thanks for catching that and fixing it in this PR!

PR #647, opened 2023-05-19
add linebreak in looping example

ndporter:
Add a newline in the print for the looping over files minimum example to improve readability of output.

### DIFF
--- a/episodes/14-looping-data-sets.md
+++ b/episodes/14-looping-data-sets.md
@@ -27,11 +27,12 @@ exercises: 10
 import pandas as pd
 for filename in ['data/gapminder_gdp_africa.csv', 'data/gapminder_gdp_asia.csv']:
     data = pd.read_csv(filename, index_col='country')
-    print(filename, data.min())
+    print(filename, '\n', data.min())
 ```
 
 ```output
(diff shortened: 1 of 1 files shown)


ndporter:
I noted later: I think there are other exercises this update would apply to as well. I won't have time to check in the near future though. Probably either someone should scan through the code/output in this episode or create an issue to follow up.

martinosorb:
Hi @ndporter, thank you for the PR. I note however that this requires explaining what `\n` means. Do you remember if there's another place before this point where it is used in the lesson materials?

Alternatively, what do you think of doing two prints? As in
```
print(filename)
print(data.min())
```

ndporter:
The `\n` construction is used as early as the[pandas data frames](
http://swcarpentry.github.io/python-novice-gapminder/aio.html#use-comparisons-to-select-data-based-on-value)
lesson (link from aio because doing search on my phone). I don't know that
they're actually introduced in the lesson anywhere, though.

So you're right that it would be good to explain them somewhere, and if not
in a main lesson then at least in instructor notes. But I was actually
following the model of earlier lessons in the PR.

PR #650, opened 2023-06-06
make sure plot elements are not truncated

### LINKED ISSUE #638: savefig truncates figure elements
In the examples for using the savefig() function on page 9 [[Please](http://swcarpentry.github.io/python-novice-gapminder/09-plotting.html)](http://swcarpentry.github.io/python-novice-gapminder/09-plotting.html), the exported figures truncate axis labels.

Is there an easy (appropriate level) modification that could be added so the full figures export? I'm new to Python and this immediately sent me away from the lesson to troubleshoot, so I think it would be good if the code in the lesson exported full figures. 

I ended up adding the argument

`bbox_inches = "tight"`

but I don't know what it means or if there is a better way to do this.

Simon-Stone:
Fixes #638 by adding calls to  `tight_layout` to the respective solution.

### DIFF
--- a/episodes/09-plotting.md
+++ b/episodes/09-plotting.md
@@ -348,7 +348,7 @@ matplotlib will make this  variable refer to a new empty figure.
 Therefore, make sure you call `plt.savefig` before the plot is displayed to
 the screen, otherwise you may find a file with an empty plot.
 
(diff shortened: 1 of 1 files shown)


martinosorb:
It looks good to me but I'd like opinions from other maintainers before merging.

alee:
I think it's a good addition as well. Only thing a bit leery of is the caveat in the docs `This is an experimental feature and may not work for some cases. It only checks the extents of ticklabels, axis labels, and titles.` if that's still true.

We could also consider changing the example code to use [constrained layouts](https://matplotlib.org/stable/tutorials/intermediate/constrainedlayout_guide.html) instead?

Issue #204, opened 2017-06-12
Remove "Local and Global Variable Use" challenge


rgaiacs:
I think the challenge "Local and Global Variable Use" at http://swcarpentry.github.io/python-novice-gapminder/15-scope/ is poor design and should be removed.

![screenshot_2017-06-12_14-48-42](https://user-images.githubusercontent.com/1506457/27036966-77db19e6-4f7e-11e7-934f-623bcf07be06.png)

vahtras:
Hi Raniere, I agree.

PR #651, opened 2023-06-07
Fix issue #204: Remove a poorly designed challenge

Simon-Stone:
Fixes issue #204 by simply removing the challenge in question.

### DIFF
--- a/episodes/17-scope.md
+++ b/episodes/17-scope.md
@@ -63,25 +63,6 @@ NameError: name 'temperature' is not defined
 
 :::::::::::::::::::::::::::::::::::::::  challenge
 
-## Local and Global Variable Use
-
-Trace the values of all variables in this program as it is executed.
-(Use '---' as the value of variables before and after they exist.)
-
-```python
-limit = 100
-
-def clip(value):
-    return min(max(0.0, value), limit)
-
-value = -22.5
-print(clip(value))
-```
-
-::::::::::::::::::::::::::::::::::::::::::::::::::
-
-:::::::::::::::::::::::::::::::::::::::  challenge
-
 ## Reading Error Messages
 
 Read the traceback below, and identify the following:


tobyhodges:
How about this?

````markdown
:::::::::::::::::::::::::::::::::::::::  challenge

## Tracing Global Variables

The code block below is a repeat of the example above, 
but with a global variable called `temperature` defined 
before the `adjust` function definition.

```python
pressure = 103.9 
temperature = 294.15 # checkpoint 1

def adjust(t):
    temperature = t * 1.43 / pressure
    return temperature # checkpoint 2

print('adjusted:', adjust(0.9))
print('temperature after call:', temperature) # checkpoint 3
```

1. Think through the code block line-by-line, 
   tracing the value of the variables as the program is run. 
   Fill in the table with the values of variables 
   called `pressure`, `temperature`, and `t` 
   at each of the checkpoints (marked with comments in the code block). 
   Write "undefined" if the value of a variable has not yet been set.

    | Checkpoint | `pressure` | `temperature` | `t` |
    | ----------| ---------- | ------------- | ----------- |
    | 1          |            |               |     |
    | 2          |            |               |     |
    | 3           |            |               |     |
1. **Before running this code block,** think about what you expect the result to be. 
   Can you identify which of the options below is correct?
    1. A `NameError` again
    1. ```output
      adjusted: 0.01238691049085659
      temperature after call: 294.15
      ```
    1. ```output
      adjusted: 0.01238691049085659
      temperature after call: 0.01238691049085659
      ```
1. Now run the code. Did you predict the result correctly? 


:::::::::::::::  solution

After the global variables are first defined, 
their values are not modified by any of the subsequent lines.
No global variable called `t` is ever created outside the `adjust` function,
so this variable remains undefined at all of the checkpoints:

| Checkpoint | `pressure` | `temperature` | `t` |
| ----------| ---------- | ------------- | ----------- |
| 1          |     103.9       |      294.15         |  undefined   |
| 2          |     103.9       |      294.15         |  undefined   |
| 3          |     103.9       |      294.15         |  undefined   |

This means that the output produced by the code block is option 2:

```output
adjusted: 0.01238691049085659
temperature after call: 294.15
```

:::::::::::::::::::::::::

::::::::::::::::::::::::::::::::::::::::::::::::::
````

PR #654, opened 2023-07-06
628 explicit df names

### LINKED ISSUE #628: improve data frame names in the lesson
Creating an issue from #470 which was not merged but will be preserved in a new feature branch for the interim

The gist is we should use better variable names for data frames than `df` and `data`, and to explicitly provide descriptive names and explain why descriptive names are better in the lesson. These improved variable names should be threaded through all the episodes as well.

sjvrijn:
Fix #628 

Episode 8:
- Renamed `df` to `europe`

Episode 16:
- Renamed `df` to `asia` for the first specific case
- Renamed `df` to `countries` in the function writing exercise, since the generic `continent` is already used

### DIFF
--- a/episodes/08-data-frames.md
+++ b/episodes/08-data-frames.md
@@ -333,7 +333,7 @@ and the Gapminder GDP data for Europe has been loaded:
 ```python
 import pandas as pd
 
-europe = pd.read_csv('data/gapminder_gdp_europe.csv', index_col='country')
+data_europe = pd.read_csv('data/gapminder_gdp_europe.csv', index_col='country')
 ```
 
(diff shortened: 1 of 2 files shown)


martinosorb:
Hi @sjvrijn, thanks for the PR. It looks good, but I have a comment: episode 07 uses `data_oceania` instead. I suggest we use `data_europe` etc. (or change the other one to `oceania`, but I like the first option better, personally).

sjvrijn:
@martinosorb good point for consistency, I've updated my changes.

martinosorb:
LGTM. Merging!

PR #655, opened 2023-07-24
Remove reference to 'two dimensional lists'

### LINKED ISSUE #491: list of lists are not two-dimensional
In the list episode in the section on appending items it says that _appending a list to a list makes the result two-dimensional_. Lists are always one-dimensional. Only arrays can be multi-dimensional. List of lists arise because a list can store any object (including a list). Sometimes lists of lists can be converted to an array if the length of the inner lists are the same.

I would probably say that lists are always one dimensional and refer to arrays for multi-dimensional data-structures.

martinosorb:
Fixes #491.


alee (approved):
Thanks for the fix @martinosorb, LGTM!

PR #668, opened 2023-11-14
Update 02-variables.md upper-lower case change

teresa-m:
Thanks a lot for the great learning materials. I have one suggestion. Within the Tutorial Plotting and Programming in Python in the Variable and Assignment part, I would change the sentence about upper and lower case naming conventions for variables. The original statement 'upper-case letters are the conventions ... so we use lower case' is ambiguous. I guess this was just mixed up, but if not this would need further explanation in my opinion.

### DIFF
--- a/episodes/02-variables.md
+++ b/episodes/02-variables.md
@@ -185,7 +185,7 @@ print(len('helium'))
 
 - Python thinks that upper- and lower-case letters are different,
   so `Name` and `name` are different variables.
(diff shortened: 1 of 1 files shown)


vahtras:
Thanks for raising the question. However, I don't think the suggested change makes anything clearer. There are a number of conventions and style recommendations which may mean going down a rabbit hole.  Any alternative formulation, anyone?

PR #672, opened 2024-02-16
Update Conditional Logic for Grade Assignment

Shivay-Shakti:
## Summary
This pull request updates the conditional logic used for assigning letter grades based on numerical scores.

## Changes
- Changed the condition for a 'C' grade to check for scores between 70 and 79 inclusive.
- Updated the condition for a 'B' grade to check for scores between 80 and 89 inclusive.

## Motivation
The previous logic incorrectly assigned a 'C' grade to any score above 70 and a 'B' grade for any score above 80, without upper bounds. This update corrects the grade assignment to reflect the standard grading scale.

## Testing
- Added test cases for the boundary conditions to ensure the grades are assigned correctly.
- Ran the entire test suite to confirm that no existing functionality is broken.

### DIFF
--- a/episodes/13-conditionals.md
+++ b/episodes/13-conditionals.md
@@ -111,16 +111,16 @@ for m in masses:
 
 ```python
 grade = 85
-if grade >= 70:
+if grade >= 70 and grade < 80:
     print('grade is C')
-elif grade >= 80:
+elif grade >= 80 and grade < 90:
     print('grade is B')
 elif grade >= 90:
     print('grade is A')
 ```
 
 ```output
-grade is C
+grade is B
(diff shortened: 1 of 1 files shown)


alee:
Thanks for taking the time to submit this PR! @sborrego submit the PR right before this that also addresses this issue in a simpler way with less conditional checks so I think we will go with that one. I'm guessing you all caught this during a carpentries workshop?

Thanks again, we appreciate your contributions!

Shivay-Shakti:
That's amazing, yes we found it during the workshop. Thanks, @sborrego and @alee for your pro-activeness!

PR #674, opened 2024-03-05
fix: clarify pandas usage with non-numeric columns

alee:
pandas now raises errors when computing the mean on a dataframe with non-numeric columns

add a note to the challenge describing the issue and provide a few additional ways of solving it

Thanks to @davidwilby for finding this bug and outlining possible solutions to it in #670


alee:
This was also addressed and fixed by @wgriffa in #666 - apologies for missing that and not crediting you with the fix as well!

PR #676, opened 2024-04-12
Workbench link in Readme

martinosorb:
The link in README still refers to the old docs for the lesson template. I have updated it to link to the Workbench docs.


alee (approved):
thank you @martinosorb !

PR #677, opened 2024-04-13
Remove the use of <div> in episode 01

martinosorb:
This is apparently needed for the lesson to be workable with Crowdin for internationalisation purposes (see [this thread on Slack](https://carpentries.slack.com/archives/CL357S0CD/p1708392714885819?thread_ts=1706754620.155089&cid=CL357S0CD)).

Please note that I can't, at the moment, check whether the version in this PR will actually work in Crowdin -- but the problem there is supposed to be about the use of \<div\>.

What the PR does is to change the [section on Markdown](http://swcarpentry.github.io/python-novice-gapminder/01-run-quit.html#the-notebook-has-command-and-edit-modes.) of the first episode to look like this:
image

Also fixes a typo by which the level-2 header looked the same as the level-1 header.

### DIFF
--- a/episodes/01-run-quit.md
+++ b/episodes/01-run-quit.md
@@ -339,161 +339,64 @@ You need to be in Command mode (Press <kbd>Esc</kbd> if your cell is blue).  Typ
 
 ### Markdown does most of what HTML does.
 
-<div class="row">
+Table: Showing some markdown syntax and its rendered output.
 
-  <div class="col-md-6" markdown="1">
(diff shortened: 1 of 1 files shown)


alee:
thanks @martinosorb ! I'm going to close this PR since it's originating from your main branch which will cause problems down the line if merged but I've cherry picked it into #678

martinosorb:
Thank you so much for the quick response and for the extra work!

PR #679, opened 2024-04-16
Refactored Episode 11 "Lists": List combinations, nested lists, removing list elements

### LINKED ISSUE #618: Use pop() instead of del in the Lists lesson?
Hi team,
I teach this sequence every term as a free workshop at the library with the assistance of an undergraduate CS major. Thank you for all your hard work to build something that works so well. While I myself have no opinion on using `pop()` vs del in Python, every single undergraduate whom I've invited teach the [Lists lesson](http://swcarpentry.github.io/python-novice-gapminder/11-lists/index.html) has refused to teach `del` and instead taught `pop()`. This makes me think that there must be a strong preference in programming for using `pop()`, and that it would be of use to our students to follow convention. If, however, this convention is particular to the U of Oregon's Computer Science Department, I'm happy to withdraw the suggestion.
If everyone agrees that `pop()` is the way to go, I'm happy to work with my students to make the PR.
Gabriele

saiboxx:
Includes and closes #618: Use pop() instead of del in the Lists lesson?

---
When working through the "Lists" episode I noticed potential for streamlining and isolating some proposed concepts.

1. For the handling of `extend`, I created a separate section for merging lists.
2. I moved the idea of nested lists to the "mixed data type list" section.
3. I generalized the section about removing items.

---

Hi Carpentry Team,

As I went through the "Lists" episode, I noticed that some programming concepts are mixed or truncated in the material.

First, the [section about append](http://swcarpentry.github.io/python-novice-gapminder/11-lists.html#appending-items-to-a-list-lengthens-it.) introduces the concepts of lists merging via `extend` and also the possibility of nested lists.
I extracted the existing part of `extend` under the general umbrella of merging lists, as I think the episode would benefit from this logical separation.
Additionally, I introduce list merging via the `+` operator, which is the pythonic way, often used in practice.

I moved the discussion of lists being able to have lists as items into the [fitting section](http://swcarpentry.github.io/python-novice-gapminder/11-lists.html#lists-may-contain-values-of-different-types.) about mixing data types. As lists are just a data type, the possibility of lists having lists as members is just a logical conclusion.

Lastly, I tried to generalize the [section about using del to remove items](http://swcarpentry.github.io/python-novice-gapminder/11-lists.html#use-del-to-remove-items-from-a-list-entirely.).
`del` is not the only option to remove list elements and `pop` is also an essential element. This was also discussed in #618.
I incorporated `pop` into the existing example with `del`.

I would be happy for feedback.
This is my first issue in this repository, so please have mercy if I have overlooked some guidelines. :)

Cheers,
Tobias

### DIFF
--- a/episodes/11-lists.md
+++ b/episodes/11-lists.md
@@ -86,29 +86,38 @@ primes has become: [2, 3, 5, 7]
   - Deliberately resembles the way we refer to things in a library.
 - We will meet other methods of lists as we go along.
   - Use `help(list)` for a preview.
-- `extend` is similar to `append`, but it allows you to combine two lists.  For example:
+
+## Multiple lists can be combined.
+- `extend` is a list *method* like `append` but it allows you to combine two lists.  For example:
 
 ```python
-teen_primes = [11, 13, 17, 19]
-middle_aged_primes = [37, 41, 43, 47]
+primes = [2, 3, 5, 7]
+more_primes = [11, 13, 17, 19]
 print('primes is currently:', primes)
-primes.extend(teen_primes)
+primes.extend(more_primes)
 print('primes has now become:', primes)
-primes.append(middle_aged_primes)
-print('primes has finally become:', primes)
 ```
 
 ```output
 primes is currently: [2, 3, 5, 7]
 primes has now become: [2, 3, 5, 7, 11, 13, 17, 19]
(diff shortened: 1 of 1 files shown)


vahtras:
Thank you for your suggestions. These are on a number of topics we would prefer to review separately. Please consider them separately and present them in the issue tracking system first before making a PR.

PR #681, opened 2024-06-11
Improve formatting of episode 16 solution

danny-lloyd:
Simulating a dynamical system solution had a new line in the code block which caused the whole block not to be rendered.

URL: https://swcarpentry.github.io/python-novice-gapminder/16-writing-functions.html

Before and after screenshots below:

Screenshot 2024-06-11 at 09 54 13

After:

Screenshot 2024-06-11 at 09 54 26

### DIFF
--- a/episodes/16-writing-functions.md
+++ b/episodes/16-writing-functions.md
@@ -573,21 +573,25 @@ density. In the model, time takes discrete values 0, 1, 2, ...
 
 ## Solution
 
-1. ```python
+1.
+  ```python
   def logistic_map(x, r):
       return r * x * (1 - x)
   ```
 
-2. ```python
+2.
+  ```python
   initial_population = 0.5
   t_final = 10
   r = 1.0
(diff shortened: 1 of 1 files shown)


vahtras:
Good catch. However there is some strange side effect on the last solution cell. Please check

Screenshot 2024-06-11 at 21 52 06

alee:
Thanks! I checked the branch out locally and rendered it using the latest sandpaper and it looked good to me, maybe there was an older version of sandpaper?

example docker-compose.yml file to be placed in the root directory and can render the site via `docker compose up -d`:

```
services:
  server:
    image: ghcr.io/uomresearchit/sandpaper:latest  # switch to `latest_arm` for macos
    volumes:
      - ./:/siteroot
    ports:
      - "127.0.0.1:4321:4321"
```

vahtras:
I did an upgrade of  sandpaper and the the problem disappeared. Who knows what is was.

PR #688, opened 2024-11-01
Changes in dataset

stacyrse:
- added a new dataset
- created new plots
- created new questions/adapted the old material

This is not the final version, I am hoping to check the layout with the new build


alee:
Hi there, did you submit this PR to the correct base repository? Please make sure you follow the contributor guidelines when submitting a PR, including a clear description of what you're attempting to modify and why. 

If you're trying to customize your own version of the gapminder lesson you'll want to adjust your PR base repository targets :sweat_smile:

You should also be careful not to commit generated files, e.g., the figures.

Cheers!

alee (commented):
Appears to be a test and includes generated files that should not be stored in the repo.

Issue #673, opened 2024-03-05
axis labels are wonky in gdp mean exercise


alee:
### How could the content be improved?

#670 exposed some issues in the rendering of the comparing data challenge

![gdp](https://github.com/swcarpentry/python-novice-gapminder/assets/5682309/af496d6d-1620-4ccf-ae27-7d9ff7c220f3)

### Which part of the content does your suggestion apply to?

https://swcarpentry.github.io/python-novice-gapminder/14-looping-data-sets.html#comparing-data

georginaalbadri:
Hello,

Two options depending on the amount of complexity we want to add. The first is just to rotate the xticklabels; this can be added prior to the plt.legend() line (outside of the loop):

`# rotate the xticklabels on the plot for better readability`
`plt.xticks(rotation=90)`
`# stop the labels being cut off when saving the figure`
`plt.tight_layout()`

![rotated_labels](https://github.com/swcarpentry/python-novice-gapminder/assets/26206247/c6770223-8ab0-4179-8809-f133b28930fc)


The second, the preferred option for a more presentable graph but likely too complex to include at this point in the tutorial is to split the strings in the xticklabels and keep only the year, since the rest of the label is the same for all the data. Add this prior to the plt.lengend() line (outside of the loop):

`# set the x-labels to be the year in the column headings of the dataframe`
`# get dataframe column headings `
`headings = dataframe.columns[1:]`
`# split the strings into two parts on the underscore and get the second part (the year)`
`years = headings.str.split('_').str.get(1)`
`# use the years as x-axis labels`
`ax.set_xticks(range(len(years)))`
`ax.set_xticklabels(years)`
`ax.set_xlabel('Year')"`

![split_labels](https://github.com/swcarpentry/python-novice-gapminder/assets/26206247/02c1a86e-1f1d-46e5-b4aa-18fcb2b12bc4)

I think as this part of the lesson is on looping over the data rather than presentation the rotation is probably sufficient!

alee:
Thanks for sharing @georginaalbadri ! A PR for this would be welcome, but I can also submit one on your behalf next week.

PR #691, opened 2024-12-03
Issue 673: Add title and just year as x-axis label

nikk-nikaznan:
Issue #673

Extract the years from the columns (so instead of `gdpPercap_<year>` will just be the year), and add title to the plot as `GDP Per Capita for Regions Over Time`.


vahtras:
Thanks. This change to the suggested solution is not directly visible in the workshop material but if one downloads the code and generates the graph one will see be a substantial improvement: avoiding ugly overlapping x-labels. There is nothing to question here so I will accept it.

PR #706, opened 2025-11-18
Update sentence in 'Plotting' lesson to improve tone & clarity

### LINKED ISSUE #705: 'Plotting' lesson: avoid bad practice of stating 'X is simple'
### How could the content be improved?

In the 'Plotting' lesson which I have taught from this week as part of an internal training course, there is a line in the section near the start under the heading 'matplotlib is the most widely used scientific plotting library in Python' which states:

> Simple plots are then (fairly) simple to create.

however it is fairly well-known and -publicised (Google via its [Developer Documentation Style Guide](https://developers.google.com/style/tone?utm_source=chatgpt.com#some-things-to-avoid-where-possible), Microsoft e.g. quoted in its older guide ['Microsoft Manual of Style'](https://jasonhall.ca/wp-content/uploads/2015/08/Microsoft.Press_.Microsoft.Manual.of_.Style_.4th.Edition.Jan_.2012.pdf?utm_source=chatgpt.com) etc.) that it is bad practice to claim that something (technical) is 'simple' or 'easy' or similar since it is subjective and what may be so to the writer(s) or those with experience could not be so to the reader and therefore frustrate them put them off, or just sound patronising. Even with the '(fairly)' adverb modifier, this is still the case.

So I/we think the statement should be rephrased. Notably in this case I think the intention is to convey that it only takes a small number of lines i.e. not much code (one main call) is required, more specifically and objectively, so the statement should be updated from asserting simplicity to asserting on conciseness (concrete suggestion below).

(To be clear, the first use of the word 'simple' is fine because it's not referring to the process of doing something that any reader could try i.e. covering a procedure, in that case it's describing an object.)

### Suggestion

In the statement quoted above, 'are then (fairly) simple' should be rephrased to something along the lines of 'can be created with just one call [to `plt.plot`]' (last bit square-bracketed bit optional, as the code snippet below it showcases this) - which is specific and objective and as per the probable intention of the writer of said line when saying more-ambiguous and subjective 'simple'.

### Which part of the content does your suggestion apply to?

http://swcarpentry.github.io/python-novice-gapminder/09-plotting.html#matplotlib-is-the-most-widely-used-scientific-plotting-library-in-python-

sadielbartholomew:
Closes #705, plus add several extra words on the updated sentence to clarify the arguments to use to `plt.plot` which in my opinion is useful in addition and otherwise only implied i.e. to be picked up by assumption from the example code snippet - strictly a separate issue so added as a separate commit, but might as well raise in this PR (can revert if controversial but doubt it will be so!).

_Please briefly summarise the changes made in the pull request, and the reason(s) for making these changes._

Update the episode/lesson 'Plotting' to avoid use of the subjective term 'simple', by replacing with the more specific and objective assertion on conciseness, with (whilst touching that sentence) a very small amount of text added to clarify what arguments `plt.plot` takes.

_If any relevant discussions have taken place elsewhere, please provide links to these._

n/a

### DIFF
--- a/episodes/09-plotting.md
+++ b/episodes/09-plotting.md
@@ -27,7 +27,7 @@ exercises: 15
 import matplotlib.pyplot as plt
 ```
 
-- Simple plots can be created with just one call, to `plt.plot` with the x variable and then y variable to
+- Simple plots can be created with a single call, to `plt.plot` with the x variable and then y variable to
   plot as the arguments:
 
 ```python


episodes/09-plotting.md
vahtras: How about "just one call" -> "a single call"?

episodes/09-plotting.md
```diff
@@ -27,7 +27,8 @@ exercises: 15
 import matplotlib.pyplot as plt
 ```
 
-- Simple plots are then (fairly) simple to create.
+- Simple plots can be created with a single call, to `plt.plot` with the x variable and then y variable to
```
alee: How about:

`Simple plots can be created with a single call to `plt.plot` with the x and y variables as arguments:`

PR #707, opened 2025-11-25
Update reflection exercise in coffee break

### LINKED ISSUE #562: update afternoon break episode 15 reflections
Currently the episode 15 reflections text expects to happen after we go over functions but due to lesson episode reordering this is no longer the case. We either need to come up with other reflections or move the break after functions (which might be too long)

https://swcarpentry.github.io/python-novice-gapminder/15-coffee/index.html

https://swcarpentry.github.io/python-novice-gapminder/16-writing-functions/index.html

JostMigenda:
In the afternoon coffee break, ensure that the reflection questions correspond to an episode before the break. Previously, they corresponded to an episode coming after the coffee break. This closes #562, see discussion there.

### DIFF
--- a/episodes/15-coffee.md
+++ b/episodes/15-coffee.md
@@ -9,10 +9,5 @@ break: 15
 
 Over break, reflect on and discuss the following:
 
-- A common refrain in software engineering is "Don't Repeat Yourself". How do the techniques we've learned in the last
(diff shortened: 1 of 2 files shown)


alee:
Thanks for the PR! I like the new text, but may preserve the original reflections in the next lesson as a preview or callout box. And we may still consider adding something about conditionals as well...

alee (approved):

